PR #3, opened 2024-10-30
feat: example of using DI, React context, and repo structure

marinovl7:
- Example PR (Do not merge without reworking it) 
- Usage of repositories, dependency injection and react context
- Compare usage of react-context with using the following: 

```typescript
export default async function Elements() {
  // Example of how to use the container to resolve a dependency
  const countryRepository = container.resolve<CountryRepository>('CountryRepository');

  const countryData = await countryRepository.getCountryData(50);

  return (
    <Suspense fallback={<div>Loading...</div>}>
      <Chart data={countryData.fcsGraph} />
    </Suspense>
  );
}
```

Perfrmance seems better for me + we are actually fetching data on the server. Using React Context requires the component to be a client component, because the Context API is a Browser API (Uses localStorage kinda). Nevertheless we should also consider request encapsulation, because the context way encapsulates the data fetching inside the chart component, which actually makes sense (not the country data but the fcsGraphData when we refactor the APIs)

Will require second opinion here, how we should do it.

### DIFF
--- a/src/components/Charts/ClientChart.tsx
+++ b/src/components/Charts/ClientChart.tsx
@@ -1,0 +1,38 @@
+'use client';
+
+import { useEffect, useState } from 'react';
+
+import { Chart } from '@/components/Charts/Chart';
+import { useCountry } from '@/domain/contexts/CountryContext';
+import { CountryData } from '@/domain/entities/country/CountryData';
+
+export function ClientChart() {
+  const { getCountryData } = useCountry();
+  const [data, setData] = useState<CountryData | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    async function fetchData() {
+      try {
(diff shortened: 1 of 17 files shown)


src/components/Charts/ClientChart.tsx
```diff
@@ -0,0 +1,38 @@
...
+export function ClientChart() {
+  const { getCountryData } = useCountry();
+  const [data, setData] = useState<CountryData | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    async function fetchData() {
+      try {
+        const countryData = await getCountryData(50);
+        setData(countryData);
```
Tschonti, line 18: maybe the context could store and distribute the returned data, so that we don't have to fetch it every time. especially if this data is not likely to change too often (and I assume it doesn't, since it's historical data. or will we have fcs data for the future, as well?

PR #9, opened 2024-11-03
Implement network calls to the backend

Tschonti:
- The data that is needed for the initial load is fetched server side (in page.tsx, which is a server component). This is useful because Next automatically caches request results and in our case this data doesn't really change. The cache TTL is set to the 12 hours for now.
- **Unfortunately, adm0data.json is 2.2MB, and Next can only cache up to 2MB for files, so right now this json file is downloaded on the server every time, so the initial load is really slow. We need to split this file in the backend code once we have access to it. Most of the data inside this file is just the borders of countries, which doesn't change, this could be static json file in the frontend codebase tbh.**
- I used Tanstack Query for the client-side network calls. It manages the states of network calls and provides client-side cache (so the data is stored in the user's browser). [More info here](https://tanstack.com/query/latest)
- Data that is only needed on demand, like additional country data, is fetched on demand. There is a useQuery hook defined for each request. When a component that uses this hook is rendered, the query is executed and it's result is stored in cache. Cache TTL is defined in the queryClient, for now it's 1 hour, can later be customized for each request if needed. If other components are rendered that use this hook, the request won't be sent, instead the data will be provided from cache.
- There's a really ugly overlay on the map to demonstrate data fetching. Its alignment will have to be figured out, but this wasn't the scope of this task. The overlay is rendered using the data that is fetched on the server. When you click on a country, a Popup is rendered that uses a hook that fetches more data about the country. The first time a country is clicked, the request is executed. On subsequent clicks, the data is returned from the cache (unless 1 hour passed since the last request). This Popup is just an example component, the sidebar elements can later use similar logic to fetch the data they need to show.


marinovl7 (approved):
Awesome job! Can't say more just amazing!

PR #15, opened 2024-11-06
Feature/f 33 chat implementation

Esoteriker:
Sorry for later PR, it took some time to refactoring the style issue: especially remove inline style and use tailwind css, it goes smoothly when bug in tailwind config fixed. 

There are some points:
- Regarding backend integration part @ahmtslmngcl  will also plan to refactoring sth, I'm not if you need do it in this branch or open a new ticket
- Typing animation is still pending, current implementation could be delete later, if text is render als markdown from backend side, we will change this component and logic
- Add some css definition into global.css related to waiting AI response animation, I hope it is not big issue.
- Since there are a lot custom style variable related to Chatbot, so a lot of customisation has also been added to the tailwind config

- Since not all icons could be found in iconsax-react, so there is still other package be used.


src/components/TypingText/TypingText.tsx
Esoteriker, line 1: This component and typing logic could be changed later depends on @ahmtslmngcl 's work.

src/domain/constant/chatbot/Chatbot.ts
Esoteriker, line 1: Correct me if constant should be put somewhere else

src/components/Chatbot/Chatbot.tsx
```diff
@@ -0,0 +1,334 @@
...
+  const [isFullScreen, setIsFullScreen] = useState(false);
+  const [isSidebarOpen, setIsSidebarOpen] = useState(false);
+  const [chats, setChats] = useState<IChat[]>([{ id: 1, title: 'Chat 1', messages: [] }]);
+  const [currentChatIndex, setCurrentChatIndex] = useState(0);
+  const [input, setInput] = useState('');
+  const [isTyping, setIsTyping] = useState(false);
+  const inputRef = useRef<HTMLTextAreaElement>(null);
+  const chatEndRef = useRef<HTMLDivElement>(null);
+  // TODO: get isMobile from context later?
+  const isMobile = useMediaQuery('(max-width: 640px)');
```
Esoteriker, line 41: Currently no idea if we will get isMobile state from context or from nextjs? It is needed in js logic.

src/services/api/chatbot.ts
Tschonti, line 1: I think we should use a repository instead of this service file here as well. So you should create an interface called ChatRepository in domain/repositories (with just two methods, testHealth and sendMessage) and a ChatRepositoryImpl in infrastructure/repositories. Then register the implementation in container.tsx so all the components can reach the instance without knowing the exact implementation.

Tschonti (commented):
generally, all types, interfaces and props should go into the domain folder, while util functions should be in the operations folder

marinovl7 (changes requested):
Overall AWESOME Job! Just had a few comments regarding code style, other than that functionality, structure, finalization everything rocks! Also generally one thing I noticed in the PR, there are no return types for your functions. Please add return types for each function 

```
const function = (): void => {...}
```

src/components/Chatbot/Chatbot.tsx
```diff
@@ -0,0 +1,334 @@
+/* eslint-disable no-nested-ternary */
+
+'use client';
+
+import { Button, Card, CardBody, CardFooter, CardHeader, Divider, Tooltip } from '@nextui-org/react';
+import clsx from 'clsx';
+import { Bot, Maximize2, Minimize2, PanelLeftClose, PanelLeftOpen, Send, X } from 'lucide-react';
```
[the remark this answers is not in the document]
bohdangarchu: pls use [iconsax-react](https://www.npmjs.com/package/iconsax-react)🙂

src/services/api/chatbot.ts
Esoteriker: Yes, @ahmtslmngcl will continue refactoring on this part 👍

src/components/Chatbot/Chatbot.tsx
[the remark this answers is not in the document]
Esoteriker: For reasons of respecting the design, some of the icons can't be found in this library so I'm using lucide-react, otherwise I'd be using iconsax-react!

src/domain/entities/chatbot/BackendCommunication.ts
ahmtslmngcl, line 1: I was not sure if APIError should be placed in here. Would be grateful for feedback!

src/components/Chatbot/Chatbot.tsx
ahmtslmngcl, line 1: wanted to rename QueryResponse "response" field to "text" so that the call would be "response.text". But interestingly renaming it (and all its occurences) caused the system to fail when trying to interact with the chatbot.

src/components/TypingText/TypingText.tsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  useEffect(() => {
+    let index = 0;
+
+    const typeCharacter = () => {
+      if (index < text.length - 1) {
+        const currentChar = text[index];
+        setDisplayedText((prev) => prev + text[index]);
+        index += 1;
```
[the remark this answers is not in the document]
Esoteriker: eslint no-plusplus, index += 1 may better

marinovl7 (approved):
Awesome job! Still few remarks and feedback to be applied, but this PR is big, so we merge it and create smaller PRs for the features and finalizations later

PR #17, opened 2024-11-08
feat: Countries with very high level of hunger alert

ahmedfarouk2000:
- There is some issues that might effect this hunger alert, this includes the request which takes a lot of time (avergae of 7 second or more) and sometimes it timed out. so I dont know if the backend will be refactored or not. This will effect the rendering data which needs to wait until request is resolved. Otherwise we might fix that by having a loading skelaton for the circle or just hide it until request is resolved (its open for discussion). 
- the style can be improved as well (open for any suggestions).
- I tried to make DataTable and Modal as generic as possible thus, it can be used if needed


src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useState, useMemo } from 'react';
+import { Table, TableHeader, TableColumn, TableBody, TableRow, TableCell, Pagination } from '@nextui-org/react';
+
+interface Column {
+  label: string;
+  key: string;
+}
+
+interface DataTableProps<T> {
```
bohdangarchu, line 9: Props interfaces should go into `/domain/props`

src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useState, useMemo } from 'react';
+import { Table, TableHeader, TableColumn, TableBody, TableRow, TableCell, Pagination } from '@nextui-org/react';
+
+interface Column {
```
bohdangarchu, line 4: All interfaces have to be in the `/domain` folder

src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+  label: string;
+  key: string;
+}
+
+interface DataTableProps<T> {
+  rows: T[];
+  columns: Column[];
+}
+
+export default function DataTable<T extends { [key: string]: any }>({ rows = [], columns }: DataTableProps<T>) {
```
bohdangarchu, line 14: generally we should avoid default values for a main parameter like rows

src/components/HungerAlert/HungerAlert.tsx
```diff
@@ -0,0 +1,64 @@
...
+import PopupModal from '@/components/PopupModal/PopupModal';
+import DataTable from '@/components/DataTable/DataTable';
+import './style.css';
+import container from '@/container';
+import { GlobalDataRepository } from '@/domain/repositories/GlobalDataRepository';
+
+export default function HungerAlert() {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [countriesWithHighHunger, setCountriesWithHighHunger] = useState<
+    { rank: number; country: string; fcs: string }[]
```
bohdangarchu, line 13: please define this separately

src/components/HungerAlert/style.css
```diff
@@ -0,0 +1,44 @@
+.pulse {
```
bohdangarchu, line 1: as already mentioned we do not use css and use tailwind instead

src/components/HungerAlert/HungerAlert.tsx
```diff
@@ -0,0 +1,64 @@
...
+            country: adm0_name,
+            fcs: `${Math.floor((fcs as number) * 100)}%`,
+          }))
+      );
+    };
+
+    fetchCountriesWithHighHunger();
+  }, []);
+
+  const columns = [
```
bohdangarchu, line 41: this should go into `operations`, you define `HungerAlertOperations` class and define a columns attribute there

src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,31 @@
+'use client';
+
+import { Modal, ModalContent, ModalHeader, ModalBody } from '@nextui-org/react';
+
+interface PopupModalProps {
```
bohdangarchu, line 5: same regarding props

src/components/HungerAlert/HungerAlert.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const fetchCountriesWithHighHunger = async () => {
+      const countryData = await container
+        .resolve<GlobalDataRepository>('GlobalDataRepository')
+        .getMapDataForCountries();
+
+      setCountriesWithHighHunger(
+        countryData.features
+          .filter(({ properties: { fcs } }) => typeof fcs === 'number' && fcs >= 0.4)
+          .sort((a, b) => (b.properties.fcs as number) - (a.properties.fcs as number))
+          .map(({ properties: { adm0_name, fcs } }, index) => ({
```
bohdangarchu, line 30: could you come up with a better name than adm0_name?

src/components/HungerAlert/HungerAlert.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [countriesWithHighHunger, setCountriesWithHighHunger] = useState<
+    { rank: number; country: string; fcs: string }[]
+  >([]);
+
+  const toggleModal = () => {
+    setIsModalOpen(!isModalOpen);
+  };
+
+  useEffect(() => {
+    const fetchCountriesWithHighHunger = async () => {
```
bohdangarchu, line 21: this should go into HungerAlertOperations

bohdangarchu (changes requested):
solid job! Please fix the linting errors otherwise we cannot merge this, also its important to follow our guidelines regarding tailwind and domain driven design, they are all listed in the README

jschoedl:
Hi, looks like you added your own table component. I was actually working with tables (but there's no PR yet because I am waiting for other stuff) - I started out making a more complex table that could be generalized into a simple table as yours. It might be good to merge both implementations later, but maybe it's okay for now to have one simpler and one more complex table in parallel, since both are using the NextUI design.

My table ([branch f-38-implement-glossary-and-methodology-pages](https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/blob/feature/f-38-implement-glossary-and-methodology-pages/src/components/Table/GroupedTable.tsx)):
Bildschirmfoto vom 2024-11-08 14-46-01
Bildschirmfoto vom 2024-11-08 14-47-37

jschoedl:
One remark regarding the styling: To me it's a bit irritating that the list modal changes its size if I reach page 3. I'd expect it to have a constant height (i.e. the height needed for 10 rows) and have more white space on the last page, maybe except for the case when there's only one page in total.

src/components/HungerAlert/HungerAlert.tsx
```diff
@@ -0,0 +1,64 @@
...
+      const countryData = await container
+        .resolve<GlobalDataRepository>('GlobalDataRepository')
+        .getMapDataForCountries();
```
Tschonti, line 24: you shouldn't download this data here, because this is already downloaded on the server in `page.tsx`. You can pass that data to this component. However, you would have to move this component to the page.tsx file, but I don't think that's a problem, since we will only have one page. Or you could fetch the data in the `layout.tsx` as well, I think next.js is smart enough not to make the same call twice.

src/components/DataTable/DataTable.tsx
ahmedfarouk2000: done

src/components/HungerAlert/style.css
ahmedfarouk2000: yeah I wish I can animate and have pseudo-elements in tailwind as well but I didnt find any.

ahmedfarouk2000:
> One remark regarding the styling: To me it's a bit irritating that the list modal changes its size if I reach page 3. I'd expect it to have a constant height (i.e. the height needed for 10 rows) and have more white space on the last page, maybe except for the case when there's only one page in total.

Yes that's true. Its annoying when navigating between pages with different hights, but I have fixed it now. @jschoedl

src/components/HungerAlert/HungerAlert.tsx
ahmedfarouk2000: The problem with this endpoint that it returns this error when its called from server side componets  =>
 **{
    "status": "resolved_model",
    "value": "{\"error\":\"Error retrieving data: Error fetching ADM0 data: ('Connection broken: IncompleteRead(1604 bytes read, 8636 more expected)', IncompleteRead(1604 bytes read, 8636 more expected))\"}",
    "reason": null,
    "_response": {
        "_bundlerConfig": null,
        "_moduleLoading": null,
        "_chunks": {},
        "_stringDecoder": {},
        "_rowState": 0,
        "_rowID": 0,
        "_rowTag": 0,
        "_rowLength": 0,
        "_buffer": []
    },
    "_debugInfo": null
}**
but when its called from client side component this error never occures. I have tried different endpoints on server side component and this issue never happens , I am not sure how exactly next.js handles those requests on different rendering components. so the work around that I have done is to pass data from layout.js (which always returns the above error) if it returns error just call the endpoint again from the client side which eventually will be resolved without an error.

Tschonti: oh I think this happens because a few days ago, their API was broken and this response was returned, and next cached it locally. When you make the request now, it returns the correct data, but it's over 2 MB so it can't update the cache with that. You can fix this by deleting the .next folder in the root dir, and then running yarn dev again.

marinovl7 (approved):
LGTM! Had few aspects that should have been fix, but that's fine. Next time please make sure to add a Skeleton Loading Indicator when you are dependent on data fetching. Nevertheless keep in mind, that we should not have two custom data tables, so align with @jschoedl there later

PR #18, opened 2024-11-09
feat: setup geojson and implement selectable countries

Lukas0912:
The main functionality of the ticket should be covered and the base for the frontend team should be there. However I ran into the following problems:

- unfortunately I had to export the colors from the theme as variables but I couldn't find a way to use tailwind classes here for each country; still the colors can be managed centrally in the tailwind config file
- the final coloring probably has to be done at a later point since in my opinion we have to use a map (only the visible map, leaflet will still be used anyway) from mapbox since it's more customizable; before that I want to get some feedback on the style of the map though
- I still couldn't find out exactly what's the exact property for clickable and non clickable countries, however this doesn't really change the base for further development


marinovl7 (approved):
Not perfect as discussed, but now the team needs this so we will repair it soon according to style and use MapBox

PR #22, opened 2024-11-11
Feature/f 30 implement map with selectable countries

Lukas0912:
Now the visible map is running over mapbox.
- Still I don't know what property is actually important for which country is clickable and which not but it should be a small change once this is known, for now it's the property 'interactive'
- Mapbox unfortunately is very isolated from our NextUI theme so I couldn't get direct access to our theme color variables working; instead I used a separate color file for the map colors as a work around, that's not a beautiful solution but the only one I managed to make it work with


Tschonti:
I think the utility `.ts` files in the `/components/Map` folder should go to the operations folder, they are not components after all.

Tschonti:
> Still I don't know what property is actually important for which country is clickable and which not but it should be a small change once this is known, for now it's the property 'interactive'
You should browse the old codebase to find out what sort of logic they use there. It'll most likely be in `WorldMap.js`

Lukas0912:
Yep, I analyzed the WorlMap file in the old code base and as I would tell the base condition for a country to be selectable seems to be the 'interactive' property that is already used in the PR. Then depending on the current mode more conditions are added for example that the needed data has to be present. All of these conditions however are very confusing and I think  they're easier to understand when we implement the modes. The base condition however that seems to be independent from the rest is 'interactive' that is relevant for this PR imo.

marinovl7 (approved):
Great Job! Once again, thanks for taking probably one of the most important and complex thing in this project and seting everything up for the others. That's how its done!

PR #25, opened 2024-11-11
Feature/f 29 Improve sidebar 

georgi4444:
- show sidebar only on home page
- global insight buttons now stay blue when selected
- new style for the page links
- next image instead of nextui image
- tooltips for alerts
- skeleton for theme switch

PR #26, opened 2024-11-11
Feature/f 38 Implement About page

jschoedl:
* I used the `@layer base` directive of tailwind for styling common elements such as h1 and p. In this case, this seemed cleaner to me than copy-pasting the same classes everywhere.
* I put the page to `/about`, but did not do anything about the routing (i.e. the sidebar does not work yet).
* The sidebar and the filters do not make sense here which will be fixed in #25 
Bildschirmfoto vom 2024-11-11 14-56-32
Bildschirmfoto vom 2024-11-11 14-56-56


marinovl7 (approved):
LGTM! Great Code Structure, great implementation, great styling! Just had to fix one thing because the styles were applied to basically everything. Next thing make sure introducing global styles doesn't affect other styles as well

PR #27, opened 2024-11-11
conflict alerts display

Tschonti:
Displays the conflicts around the world when that alert type is active

Things to note:
- The json file that stores the ~5000 conflicts is almost 1MB, which takes some time to load, maybe we need a loading indicator
- The result is cached locally, so it's faster on the second time, but still quite long as leaflet has to draw a bunch of markers (somehow this is way faster in the old map, might spend some time researching this in the next few days)
- On max zoom, there is no clustering, I think this generally looks better. However, it's a bit weird around Israel: 
image
- I didn't implement the legend, since the generic component is not ready yet


marinovl7 (approved):
Generally I'd say its a good job! However as you mentioned the rendering is expensive the first time and takes longer. We need to find a solution for that. Also please merge the new main status and add a legend so we have this completed, the legend component is already merged!

marinovl7 (changes requested):
Overal good job! We need to find a solution for the optimizations! Can you please add legend to the alerts, since the legends PR is already merged so we can roll out this feature without the loading state only for now!

Tschonti:
I added the legend. The description tooltip doesn't support markdown and doesn't look good with long text, so I just used the first paragraph from the current version.

Also implemented a very basic loading state: a spinner shows up next to the legend title. We can rollback this if you don't like it, but I think it's better than nothing.

The latency is weird, it doesn't seem to be leaflet that's slow, it's the sidebar context, but I have no idea why...

and unfortunately the zooming is quite ugly with the markers and mapbox, not sure what we can do with this...

marinovl7 (approved):
LGTM! Let's discuss the loading indicator + the zooming functionality and the sidebar context later in the meeting. It's not perfect yet, but its understandable tho, this is the first alert feature, so thank for providing that! Awesome job!

PR #30, opened 2024-11-13
Feature/f 61 finish chatbot

Esoteriker:
Finalization:

- icon size become bigger
- no opacity of background
- same dark and light color of trigger icon like sidebar, to make design consistent
- use shadow component from nextui to show chat message (not using browser scrollbar)
- improve padding and margin 
- fix issue, when Bot typing, then dont show waiting response animation
- fix scroll issue, when no content extend parent height in chat interface, then dont scroll automatic


marinovl7 (approved):
LGTM!

PR #31, opened 2024-11-14
feat: implement fcs legend and overlay

Lukas0912:
- For now I left out the functionality of the FCS map coloring since there is still the CORS error occurring, locally however this version would run if the missing header of the requested data is added
- With this PR the frontend team should also have a common basis to develop the other modes so that later there are less merge conflicts


marinovl7 (approved):
Great Job!

PR #32, opened 2024-11-14
Feature/f 68 check and replace tooltips

plaume8:
- replaced all standard "NextUI" tooltips with our custom one
- for the PdfViewer component I did not replace the usage of the NextUI tooltip -> cause here the NextUI tooltip was used to diayplay a button on request, which is not supported by our custom "basic" tooltip component which can only desplay a title and text


bohdangarchu (approved):
lgtm!

marinovl7 (approved):

PR #33, opened 2024-11-16
Hazard alerts display

Tschonti:
I only clustered the same type of hazards together, so there are still quite a lot on the map, but I think it wouldn't make sense to cluster everything together.
This PR also includes a slight rework of the PointLegend component, now a custom component can be passed to it, and it will render that for every legend item.


src/components/Map/Alerts/HazardLayer.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function HazardLayer() {
+  const { data, isPending } = useHazardQuery();
+  const hazardsByType = useMemo(() => HazardOperations.sortHazardsByType(data), [data]);
+
+  return (
+    <>
+      <div className="absolute bottom-6 right-8 z-9999">
+        <LegendContainer loading={isPending || !data} items={HazardOperations.generateHazardLegend()} />
+      </div>
```
Tschonti, line 20: this is overridden by other legend containers, so this won't be the final solution, but at least it's testable this way

marinovl7 (approved):
An example of job well done! You rock!

PR #34, opened 2024-11-16
Split up SidebarContext to fix slow state change with the alerts map

Tschonti:
Previously, whenever the state of the sidebar changed (e.g it was opened or closed), the current alert layer was rerendered, because it used the `selectedAlert` property of the same `SidebarContext`. Not sure if it's normal that this rerender took so long, but I could solve the real issue by splitting up the context into three contexts. Changing the sidebar state had no effect on the selected map and alert (and vice-versa), so we lose nothing but now the components are only rerendered when needed.

### DIFF
--- a/src/app/providers.tsx
+++ b/src/app/providers.tsx
@@ -6,6 +6,8 @@ import { ThemeProvider as NextThemesProvider } from 'next-themes';
 import type { ThemeProviderProps } from 'next-themes/dist/types.d.ts';
 import * as React from 'react';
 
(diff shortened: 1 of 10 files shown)


georgi4444:
Nice one! Really good idea and it fixed the issue

georgi4444 (approved):
LGTM

marinovl7 (approved):
Awesome job! Is good to have the three context as well to separate the concerns and Improve performance! LGTM

PR #35, opened 2024-11-17
Feature/f 51 ipc map

ahmedfarouk2000:
**The IPC Map with popOver on hovered colored countries** 
image

- I didn't use any layers in here since its only about coloring the IPCs countries.
- generateColorMap func: the logic looks complicated but its becase the IPCs data from endpoint that ends with ipc.json does not include all the data needed thus I need to get them from countryMapData endpoint that ends with adm0data.json, so I think its better to refactor those endpoints.


src/operations/map/MapOperations.ts
```diff
@@ -9,14 +10,40 @@ import { GlobalInsight } from '@/domain/enums/GlobalInsight.ts';
...
 export class MapOperations {
   static createMapboxMap(
     isDark: boolean,
-    { countries }: MapProps,
+    { countries, ipcData, selectedMapType }: MapProps,
     mapContainer: RefObject<HTMLDivElement>
   ): mapboxgl.Map {
     const mapColors: MapColorsType = getColors(isDark);
 
+    const getCountryFillPaint = (): DataDrivenPropertyValueSpecification<string> => {
```
jschoedl, line 23: maybe `getCountryFillColor`?

src/operations/map/MapOperations.ts
```diff
@@ -86,44 +107,97 @@ export class MapOperations {
...
-    baseMap.on('mousemove', 'country-fills', (e) => {
-      if (e.features && e.features.length > 0 && (e.features[0] as unknown as CountryMapData).properties.interactive) {
-        if (hoveredPolygonId) {
-          baseMap.setFeatureState({ source: 'countries', id: hoveredPolygonId }, { hover: false });
-        }
-        hoveredPolygonId = e.features[0].id;
-        if (hoveredPolygonId) {
-          baseMap.setFeatureState({ source: 'countries', id: hoveredPolygonId }, { hover: true });
+    baseMap.on('mousemove', 'country-fills', (e: any) => {
+      const feature = e.features && (e.features[0] as CountryMapData);
```
jschoedl, line 118: maybe `countryData` instead of `feature`, see above

.env_template
jschoedl, line 1: Don't delete this file! (Maybe happened accidentally when renaming it)

src/operations/map/MapOperations.ts
```diff
@@ -86,44 +107,97 @@ export class MapOperations {
...
+          const ipcColorMap = IPCMapOperations.generateColorMap(ipcData!, countries!);
+
+          if (!ipcColorMap[countryName]) {
+            popup.remove();
+            return;
+          }
+
+          const { date_of_analysis, analysis_period, ipc_percent, ipc_popnmbr } = IPCMapOperations.findIpcData(
+            countryName,
+            ipcData!
```
jschoedl, line 149: `!` doesn't fit `MapProps` (like above)

src/operations/map/MapOperations.ts
```diff
@@ -86,44 +107,97 @@ export class MapOperations {
...
+          );
+          const formattedPopNum = ipc_popnmbr.toLocaleString('en-US', {
+            useGrouping: true,
+          });
+          const title = `${countryName}`;
```
jschoedl, line 154: This line does nothing but converting `countryName` to a string, I think you can remove it and use `countryName` directly.

src/operations/map/MapOperations.ts
```diff
@@ -86,44 +107,97 @@ export class MapOperations {
...
 
+  static generatePopupContent(title?: string, body?: string, conclusion?: string): string {
+    return `
+      <div class="p-4 bg-white dark:bg-black text-black dark:text-white rounded-medium flex flex-col gap-2">
+        <h1 class="text-md font-semibold">${title}</h1>
+        <p class="text-xs">${body}</p>
+        <p class="text-xs">${conclusion}</p>
+      </div>
+    `;
+  }
```
jschoedl, line 199: Maybe that's just a personal opinion, but I'd inline this function as it is used only once and does nothing complicated.

src/styles/globals.css
jschoedl, line 1: Hm, I wonder why your CSS formatting is that different...

tailwind.config.js
```diff
@@ -134,6 +134,9 @@ module.exports = {
             fcsGreen: '#345d34',
             fcsOrange: '#ea6a2c',
             fcsRed: '#fa190e',
+            ipcStartColor: '#F6D1C1',
+            ipcMiddleColor: '#FB7453',
+            ipcEndColor: '#710013',
```
jschoedl, line 139: `ipsStart`, `ipsMiddle` and `ipsEnd` should be enough - none of the colors in this file have the suffix `color`.

jschoedl (changes requested):
LGTM! Although I haven't worked with Mapbox yet and can't say too much about it.

I added some suggestions regarding types and variable names.

src/app/page.tsx
```diff
@@ -11,13 +11,19 @@ export default async function Home() {
...
   const countryMapDataPromise = globalRepo.getMapDataForCountries();
   const disputedAreasPromise = globalRepo.getDisputedAreas();
-  const [countryMapData, disputedAreas] = await Promise.all([countryMapDataPromise, disputedAreasPromise]);
+  const ipcDataPromise = globalRepo.getIpcData();
+  const [countryMapData, disputedAreas, ipcData] = await Promise.all([
+    countryMapDataPromise,
+    disputedAreasPromise,
+    ipcDataPromise,
+  ]);
+
```
marinovl7, line 20: @Tschonti we should not fetch the IPC Data here right? Any suggestions where to move this, so that we fetch the data only on sidebar click on the ipc icon

src/components/Map/Map.tsx
```diff
@@ -24,7 +24,7 @@ export default function Map({ countries, disputedAreas }: MapProps) {
       style={{ height: '100%', width: '100%', zIndex: 1 }}
     >
       <AlertContainer />
-      {countries && <VectorTileLayer countries={countries} disputedAreas={disputedAreas} />}
+      {countries && <VectorTileLayer countries={countries} disputedAreas={disputedAreas} ipcData={ipcData} />}
```
marinovl7, line 27: maybe a new layer would be needed for the normal VectorTile layer instead? @Tschonti

src/operations/map/MapOperations.ts
marinovl7: also categorize it as `getCountryFillColorForIPC`

src/operations/map/MapOperations.ts
```diff
@@ -86,44 +107,97 @@ export class MapOperations {
...
-        if (hoveredPolygonId) {
-          baseMap.setFeatureState({ source: 'countries', id: hoveredPolygonId }, { hover: true });
+    baseMap.on('mousemove', 'country-fills', (e: any) => {
+      const feature = e.features && (e.features[0] as CountryMapData);
+
+      if (hoveredPolygonId && (!feature || hoveredPolygonId !== feature.id)) {
+        baseMap.setFeatureState({ source: 'countries', id: hoveredPolygonId }, { hover: false });
+        hoveredPolygonId = undefined;
+        baseMap.getCanvas().style.cursor = '';
+      }
```
marinovl7, line 124: When you hover a country the country color is set to nothing, this needs to be fixed, maybe that is causing the issue

src/operations/map/MapOperations.ts
```diff
@@ -86,44 +107,97 @@ export class MapOperations {
...
+          const formattedPopNum = ipc_popnmbr.toLocaleString('en-US', {
+            useGrouping: true,
+          });
+          const title = `${countryName}`;
+          const body = `Date of analysis: ${date_of_analysis} <br> Validity period: ${analysis_period}`;
+          const conclusion = `${formattedPopNum} people in IPC/CH Phase 3 and above <br> (${ipc_percent}% of people in the analyzed areas)`;
+
+          const popupContent = this.generatePopupContent(title, body, conclusion);
+          popup.setLngLat(e.lngLat).setHTML(popupContent).addTo(baseMap);
+          break;
```
marinovl7, line 160: Can't we use our tooltip here? Maybe align with @ArmanpreetGhotra

src/operations/map/MapOperations.ts
marinovl7: That's the better way to do it like it already is, to separate the concerns. Maybe a precise naming here would be not bad

marinovl7 (changes requested):
Please revisit the data fetching issues with @Tschonti and communicate about the tooltip/popup with @ArmanpreetGhotra. Precise naming for the functions is needed to distinguish between ipc functions, nutrition functions etc. Overall the request and the functionality are perfect!

src/app/page.tsx
Tschonti: Yeah it'd be great if we didn't fetch it here, but it was done so the mapbox layers and interactivity can be set up with this data. I'm not sure if mapbox is okay with dynamically creating a new layer (though it probably is).

the IPC data endpoint is ~21 KB, which is not that small, but way smaller than adm0 for example, so I don't think it makes a big difference to prefetch this.

src/styles/globals.css
ahmedfarouk2000: I think different versions might have some slight different formatting

src/operations/map/MapOperations.ts
ahmedfarouk2000: done

src/operations/map/MapOperations.ts
ahmedfarouk2000: yeah its kinda redundant becuses previously I was showing data like => **country: Germany** but since it was better to show just the country name I forgot to delete this var

src/operations/map/MapOperations.ts
ahmedfarouk2000: yeah I think I will update `MapProps` interface since `ipcData` is used in every defined `MapProps` thus it does not make sense to make it optional anymore, this will definitely fix non-null assertion issue everywhere

src/operations/map/MapOperations.ts
ahmedfarouk2000: yeah this is definitely more descriptive

src/operations/IPC/IpcMapOperations.tsx
```diff
@@ -0,0 +1,39 @@
...
+    });
+    const colorMap: Record<string, string> = {};
+    mapData.features.forEach((feature: CountryMapData) => {
+      const ipcDataForFeature = ipcDataById[feature.properties.adm0_id];
+
+      if (ipcDataForFeature) {
+        const countryName = feature.properties.adm0_name;
+        const { ipcPopulation } = feature.properties;
+
+        colorMap[countryName] = IPCMapOperations.fillIpcMap(ipcPopulation!);
```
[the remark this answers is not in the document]
ahmedfarouk2000: yeah this make sense, I didnt notice `ipcPopulation` was optional, I think once the endpoints are refactors we will remove this field anyways.

.env_template
ahmedfarouk2000: yeah might be

src/components/Map/Map.tsx
ahmedfarouk2000: I tried not to have additional layer, since having additional layer will have some sort of unnatural behaviour when zooming in or out (not sure if it's solvable or not)

ahmedfarouk2000: I think this will depend on **IPC/CH Country View**  if this will show details about country ipc phases thus, its better to have a seprate layer.

src/operations/map/MapOperations.ts
ahmedfarouk2000: We can for sure, but I prefer using the mapBox pop up which adaptively adjusts its position based on the mouse's location. e.g if there is no space on the right side of the screen, the popup will automatically appear on the left, and so on.

ahmedfarouk2000: image

ahmedfarouk2000: image

src/operations/map/MapOperations.ts
ahmedfarouk2000: what issue exactly?

src/operations/map/MapOperations.ts
```diff
@@ -115,47 +115,50 @@ export class MapOperations {
...
             useGrouping: true,
           });
-          const title = `${countryName}`;
-          const body = `Date of analysis: ${date_of_analysis} <br> Validity period: ${analysis_period}`;
-          const conclusion = `${formattedPopNum} people in IPC/CH Phase 3 and above <br> (${ipc_percent}% of people in the analyzed areas)`;
-
-          const popupContent = this.generatePopupContent(title, body, conclusion);
+          const popupDetails = `Date of analysis: ${date_of_analysis} <br> Validity period: ${analysis_period}`;
+          const popupSummary = `
+          <span class="text-base font-semibold" style="color: #E08004;">${formattedPopNum}</span>
```
jschoedl, line 158: Avoid using the style attribute, use Tailwind instead. Just add the color to `tailwind.config.js` and add the class `text-yourAddedColorName`. (same below)

And I wonder why you are using a template string here instead of a React Element - is there a reason for this? I think removing the backticks and dollar signs and replacing `class` with `className` would do the same in a more React-like way.

jschoedl (commented):
One small remark - the rest looks fine for me. :)

src/operations/map/MapOperations.ts
ahmedfarouk2000: At first, I thought about that, but since the file was `.ts`, I couldn't append React components. I completely forgot to fix that, but now I have changed the file to `.tsx` thus I can safely append it.

Tschonti:
I merged main - the MapOperations file changed a lot, so there was a huge merge conflict and had to refactor this solution to make it work. The only thing that's different than before the merge is that the inactive countries are still greyed out on the IPC map.

jschoedl (approved):

marinovl7 (approved):

PR #37, opened 2024-11-18
feature: f-56-search-for-country-selection

ahmtslmngcl:
- implemented zoom functionality to country from sidebar via search or click on the map
- used context for managing the selected country
- added dynamic zooming via using MapBox API which allowed to get GeoJSON of a selected country

Draft: 
- there is a bug when dragging and moving the mouse after the zoom has run. I didn't managed to fix it. The synchronization between mapbox and leaflet could cause it.


src/operations/map/MapOperations.ts
```diff
@@ -1,15 +1,21 @@
...
 import { CountryMapData } from '@/domain/entities/country/CountryMapData.ts';
+import { SelectedCountry } from '@/domain/entities/country/SelectedCountry';
 import { MapColorsType } from '@/domain/entities/map/MapColorsType.ts';
 import { GlobalInsight } from '@/domain/enums/GlobalInsight.ts';
 import { MapProps } from '@/domain/props/MapProps';
+import CountryRepository from '@/domain/repositories/CountryRepository';
 import { getColors } from '@/styles/MapColors.ts';
 
 export class MapOperations {
+  static countryRepository: CountryRepository = container.resolve<CountryRepository>('CountryRepository');
```
ahmtslmngcl, line 17: Is there a better way to provide the countryRepository?

src/operations/sidebar/SidebarOperations.ts
```diff
@@ -63,4 +65,19 @@ export class SidebarOperations {
...
+        name: country.adm0_name,
+        coordinates: {
+          longitude: country.longitude,
+          latitude: country.latitude,
+        },
+      }));
+    } catch {
+      return [];
+    }
+  }
```
Tschonti, line 82: I think this request is not necessary. The `getMapDataForCountries` method of the global repo is already called on the server, this data includes the countries and their centroids. For now, you can get access to this data by passing it through some components as a prop, but soon a new hook or context can be used which should be much easier.

Also, there are countries that are selectable but don't have ipc data

src/infrastructure/repositories/CountryRepositoryImpl.ts
```diff
@@ -36,4 +38,13 @@ export default class CountryRepositoryImpl implements CountryRepository {
...
   }
+
+  async getCountryGeoJSON(countryName: string): Promise<FeatureCollection> {
+    const mapboxToken = process.env.NEXT_PUBLIC_MAPBOX_ACCESS_TOKEN;
+    const response = await fetch(
+      `https://api.mapbox.com/geocoding/v5/mapbox.places/${countryName}.json?access_token=${mapboxToken}&types=country`
+    );
+    const data = await response.json();
+    return data.features[0];
+  }
```
Tschonti, line 49: we also have the coordinates for the borders of every country, so we might not need this. Of course, that's a lot more data than just the bounding box, but maybe turf could create the bounding box from that as well?

src/components/Sidebar/Sidebar.tsx
```diff
@@ -48,7 +72,19 @@ export function Sidebar() {
...
+            className="w-full"
+            variant="faded"
+            color="primary"
+          >
+            {countryData.map((country) => (
+              <SelectItem key={country.name} value={country.name}>
+                {country.name}
+              </SelectItem>
+            ))}
+          </Select>
```
Tschonti, line 87: I think it would be better if we used autocomplete here: https://nextui.org/docs/components/autocomplete

Tschonti (commented):
overall nice work, but I think we need a few modifications.
Also there should be a way to zoom out to global view once you clicked on a country

Tschonti:
the alerts stay at the same place when you click on a country, they're not affected by the zoom :/

Tschonti:
one more find: if you click on a country with the fcs map, then choose another country from the dropdown, the region layer stays on top. the selectedCountryId context should be cleared when a new country is selected. I gues this branch already had a similar context, so maybe those should be merged

Lukas0912:
The zoom to the country view is working better now, sometimes the country data doesn't adjust if you zoom out afterwards though. Also I had to completely change the clickable tiles from the nutrition branch to mapbox because by now we're to deep in mapbox. There are still some bugs but fixing them unfortunately is a very slow process since it's hard to find out how leaflet and mapbox update each other.

PR #38, opened 2024-11-18
feat: refactor map legends

Lukas0912:
- The legend components were refactored so that now all the legends shown in the map view are based on one legend container and the right data is set dynamically
- I had to split some parts of the hazardOperations file since the leaflet library is imported there and thus the component is fully rendered on client side whereas the MapLegend can partly be prerendered on server side; Otherwise there would've been an error


marinovl7 (approved):
LGTM!

PR #39, opened 2024-11-18
Feature/f 65 move high hunger countries

ArmanpreetGhotra:
Redesign the hunger alert button by making it small and removing the text from it. For the moment, don't know if it should be visible all the time or not, so if someone is collapsing the sidebar, the alert is hidden. There is kind of warning color pulse 
Screenshot 2024-11-18 at 18 37 16


marinovl7 (changes requested):
Looks great! But can you also adjust the Skeleton Loading Animation Position and Size?

marinovl7 (approved):
LGTM!

PR #40, opened 2024-11-18
diasable leaflet zoom animation

Tschonti:
This should not be our final solution to the zooming problem, but at least this way the map and the markers are always in sync, even during the zoom.


marinovl7 (approved):
LGTM! Looks super ugly, but works for now! Will be fixed this week I guess!

PR #41, opened 2024-11-18
feat: implement rainfall and vegetation maps

jschoedl:
Due to the CORS of the WFP APIs, the map doesn't work in the demo. During development I used a Plug-In that temporarily disabled the CORS guidelines.
Bildschirmfoto vom 2024-11-18 20-56-33
Bildschirmfoto vom 2024-11-18 20-56-47
Bildschirmfoto vom 2024-11-18 20-57-04
Bildschirmfoto vom 2024-11-18 20-56-56

### DIFF
--- a/src/operations/map/MapOperations.ts
+++ b/src/operations/map/MapOperations.ts
@@ -39,30 +39,26 @@ export class MapOperations {
           {
             id: 'ocean',
             type: 'background',
-            paint: {
-              'background-color': mapColors.ocean,
-            },
+            paint: { 'background-color': mapColors.ocean },
           },
           {
             id: 'countries-base',
             type: 'fill',
             source: 'countries',
-            layout: {},
-            paint: {
-              'fill-color': [
-                'case',
-                ['boolean', ['coalesce', ['get', 'interactive'], false]],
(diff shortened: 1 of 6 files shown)


marinovl7 (approved):
LGTM for me! @Lukas0912 will do the final approval and merge!

src/operations/map/MapOperations.ts
```diff
@@ -57,6 +58,16 @@ export class MapOperations {
               'fill-opacity': ['case', ['boolean', ['feature-state', 'hover'], false], 0.7, 1],
```
Lukas0912, line 58: I think this line doesn't have any effect anymore now

Lukas0912 (changes requested):
Very nice improvement of the layer system from before, I would just suggest to add another layer on top with the deactivated non interactive countries and change the base layer to a uniformly colored base layer. Right now the differentiation of selectable and non selectable countries is not possible anymore since the overlay isn't transparent.

jschoedl:
Bildschirmfoto vom 2024-11-19 10-26-40
Bildschirmfoto vom 2024-11-19 10-26-48

PR #42, opened 2024-11-18
fix: change padding and text size of popup modal

Esoteriker:
small changes regarding padding and text size of popup modal espcially in mobile version, to save more space


marinovl7 (approved):
LGTM

PR #45, opened 2024-11-19
feat: skeleton of map while loading

Lukas0912:
- I added a MapSkeleton based on SVG data in the colors of the map
- also I moved the other Map data and the data for the skeleton in a common folder in the constant area
- the loading skeleton can also be seen on the elements demo page


Tschonti:
that is super cool, love it!

marinovl7 (approved):
Great solution! I think something better could not be done, didn't even think about that either, great jib @Lukas0912

PR #46, opened 2024-11-19
use content1 as bg color everywhere

Tschonti:
alertMenu (only when the sidebar is closed), hunger alert, all accordions and chatbot opener button changed (to the bg color of the sidebar)


marinovl7 (approved):
LGTM!

PR #47, opened 2024-11-19
Avoid reloading the map when the theme or selected map changes

Tschonti:
I know I didn't have this task, but I realized this might be possible, and afaik, we didn't assign this to anyone, since we weren't sure it was possible. Well, it is, and it's not even that painful. We just have to avoid recreating the whole map when some inner state changes, instead we have to apply the change on the current map.


Lukas0912:
Wow, works great!

Lukas0912 (approved):
Wow, works great!

PR #36, opened 2024-11-17
Feature/f 64 charts option improvements

plaume8:
The main goal was to implement:
https://linear.app/world-food-programme-wfp/issue/F-64/charts-option-improvements

In addition, all the requested changes from the last pull request have been implemented: https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/21
Except for the task to “chart modal to own component”  -> I still have a question before making the changes: Wouldn't it make more sense to keep the modal inside since the regular LineChart box component and the modal share many dependencies (e.g. all the state manipulations)?"

--- 
And... I have a quick question: as requested, I moved all 'const button = <...>' components directly into the return statement. Is it generally intended for the future that JSX code should only appear within the return statement?

PR #48, opened 2024-11-20
Feature/f 52 current food consumption country view

bohdangarchu:
- links to Q&A will need to be added later
- if no data is available an empty graph is shown
- will make linecharts transparent when #36 is merged
Some examples:
image
image


src/components/Map/FcsAccordion.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { FcsAccordionOperations } from '@/operations/map/FcsAccordionOperations';
+import { cardsWrapperClass } from '@/utils/primitives';
+
+import CustomAccordion from '../Accordions/Accordion';
+import CustomCard from '../Cards/Card';
+import { LineChart } from '../Charts/LineChart';
+
+export default function FcsAccordion({ countryData, loading, countryIso3Data }: FcsAccordionProps) {
+  const deltaOneMonth = countryData?.fcsMinus1 ? countryData.fcs - countryData.fcsMinus1 : undefined;
+  const deltaThreeMonth = countryData?.fcsMinus3 ? countryData.fcs - countryData.fcsMinus3 : undefined;
```
marinovl7, line 13: useMemo here

src/components/Map/FcsChoropleth.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, [data]);
+
+  return (
+    <div>
+      <GeoJSON
+        ref={(instance) => {
+          geoJsonRef.current = instance;
+        }}
+        data={data}
+        style={() => style}
```
marinovl7, line 41: why a function?

src/components/Map/FcsRegionTooltip.tsx
marinovl7, line 1: If the value is N/A there is no need to show the text, show something like no Data Available about ...

image

src/components/Map/Map.tsx
```diff
@@ -17,8 +28,8 @@ export default function Map({ countries, disputedAreas }: MapProps) {
         [-90, -180],
         [90, 180],
       ]}
-      minZoom={MAP_MIN_ZOOM}
-      maxZoom={MAP_MAX_ZOOM}
+      minZoom={2}
+      maxZoom={18}
```
marinovl7, line 32: use again the original ones

src/components/Map/Map.tsx
```diff
@@ -1,14 +1,25 @@
...
+import FcsChoropleth from './FcsChoropleth';
 import VectorTileLayer from './VectorTileLayer';
 
 export default function Map({ countries, disputedAreas }: MapProps) {
+  const { selectedMapType } = useSelectedMap();
+  const [selectedCountryId, setSelectedCountryId] = useState<number | undefined>();
+
+  const countryStyle: L.PathOptions = {
+    color: undefined,
+  };
```
marinovl7, line 21: operations

marinovl7 (changes requested):
LGTM! Some functionality feedback still to come + some code style issues, but just a few. Generally I noticed a bug where if I click on egypt then Lybia then Egypt again there are no tooltips shown instead when I click on a region the loading animation is shown in the accordion. Also if we have alerts activated we should deactivate them, so not only the active map layer, but also the alerts. 

Small note, selectCountryId could be reworked in the future.

bohdangarchu:
New changes:
- fix code style issues 
- fix bug with going back from country to country
- deactivate alerts and current map layer - @Tschonti I created a new context for map visibility, please check
- deactivate country view and reactivate current map layer on zoom out
- display a corresponding text in accordion when no data is available

src/components/Map/FcsAccordion.tsx
bohdangarchu: can we avoid using useMemo for simple number subtraction?

bohdangarchu:
hmm I still see the bug with going back from country to country in the preview, although locally everything works... not sure why this can happen but I'll look into it

Tschonti:
the context looks good to me, a few questions/findings:
- there's no hover animation on the countries, regions in this version, though this might be because of an older master, there were a few changes with the mapbox map recently
- on a few countries (for example Niger), there's no popover, is that intentional? Is the data missing there?

bohdangarchu:
> hmm I still see the bug with going back from country to country in the preview, although locally everything works... not sure why this can happen but I'll look into it

should be fixed now

bohdangarchu:
> * there's no hover animation on the countries, regions in this version, though this might be because of an older master, there were a few changes with the mapbox map recently

should be fixed now

> * on a few countries (for example Niger), there's no popover, is that intentional? Is the data missing there?

should be also fixed now, I filter out countries without fcs data

src/components/Map/Map.tsx
```diff
@@ -27,6 +45,24 @@ export default function Map({ countries, disputedAreas }: MapProps) {
     >
       <AlertContainer countries={countries} />
       {countries && <VectorTileLayer countries={countries} disputedAreas={disputedAreas} />}
+      {selectedMapType === GlobalInsight.FOOD &&
+        countries.features &&
+        countries.features
+          .filter((countryData) => countryData.properties.interactive)
+          .filter((countryData) => countryData.properties.fcs !== null)
+          .map((countryData) => (
```
Tschonti, line 53: maybe not the scope of this task, but we should use this condition on the global map to display countries as interactive or disabled. Maybe we could look into whether we can change which countries are clickable on each map, because I assume it will be different on the IPC map for example

Tschonti (approved):

Tschonti:
oh and please merge main again to resolve the conflicts

src/components/Map/Map.tsx
bohdangarchu: yeah, we could do this as a separate task, once for all map types. Here is the [link](https://linear.app/world-food-programme-wfp/issue/F-89/clickable-countries-in-global-view)

PR #49, opened 2024-11-20
Feature/f 53 nutrition country view

ArmanpreetGhotra:
-> Implement Complete Nutrition View


src/app/Nutrition/NutritionLayer.tsx
```diff
@@ -0,0 +1,108 @@
...
+}: NutritionChoroplethProps) {
+  const [regionData, setRegionData] = useState<GeoJSON.FeatureCollection | null>(null);
+  const [showStateChoropleth, setShowStateChoropleth] = useState<boolean>(false);
+  const [selectedCountryId, setSelectedCountryId] = useState<GeoJSON.FeatureCollection | undefined>(undefined);
+
+  const geoJsonRef = useRef<L.GeoJSON | null>(null);
+
+  const map = useMap();
+
+  const handleCountryClick = async (feature: Feature<Geometry, GeoJsonProperties>, bounds: L.LatLngBounds) => {
```
bohdangarchu, line 33: this needs to go in NutritionLayerOperations

src/app/Nutrition/NutritionLayer.tsx
bohdangarchu, line 1: please call the file according to component name

src/app/Nutrition/NutritionLayer.tsx
```diff
@@ -0,0 +1,108 @@
...
+          map.fitBounds(bounds);
+        }
+      } catch (error) {
+        console.error('Error fetching country data:', error);
+      }
+    }
+    if (handleClick) handleClick(feature, bounds, map);
+  };
+
+  const features = (feature: Feature<Geometry, GeoJsonProperties>, layer: L.Layer) => {
```
bohdangarchu, line 55: operations

src/app/Nutrition/NutritionLayer.tsx
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    if (onEachFeature) onEachFeature(feature, layer);
+
+    const content = tooltip ? tooltip.render(feature) : feature.properties?.adm0_name;
+    const div = document.createElement('div');
+    const root = createRoot(div);
+    root.render(content);
+
+    layer.bindTooltip(div, { sticky: true, className: tooltip ? tooltip.className : 'countryTooltip' });
```
bohdangarchu, line 69: I think country tooltip is out of scope for this task

src/app/Nutrition/NutritionLayer.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <>
+      <GeoJSON ref={geoJsonRef} style={style} data={data} onEachFeature={features} />
+      <div className="absolute bottom-5 right-0 z-50 pr-10">
+        <LegendContainer items={[legendItem]} />
+      </div>
+      {showStateChoropleth && regionData && (
+        <StateChoropleth
+          regionNutri={selectedCountryId}
+          regionData={regionData}
+          handleClick={(feature: GeoJsonProperties) => console.log('State clicked:', feature)}
```
bohdangarchu, line 99: unneded log

src/app/Nutrition/NutritionLayer.tsx
```diff
@@ -0,0 +1,108 @@
...
+          tooltip={{
```
bohdangarchu, line 100: state tooltip also not needed here

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+import { Layer, LeafletMouseEvent, PathOptions } from 'leaflet';
+import React, { useEffect, useRef, useState } from 'react';
+import { GeoJSON } from 'react-leaflet';
+
+import CustomAccordion from '@/components/Accordions/Accordion';
+import LegendContainer from '@/components/Legend/LegendContainer';
+import { GradientLegendContainerItem } from '@/domain/props/GradientLegendContainerItem';
+import NutritionStateChoroplethProps from '@/domain/props/NutritionStateProps';
+
+interface NutrientOption {
```
bohdangarchu, line 12: this should be in domain

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+  label: string;
+  key: string;
+}
+
+interface AccordionItem {
```
bohdangarchu, line 17: domain

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+  title: string;
+  iconSrc: string;
+  description: string;
+  content: JSX.Element;
+}
+
+type LayerWithFeature = Layer & {
+  feature?: GeoJSON.Feature<Geometry, GeoJsonProperties>;
+};
+const formatNutrientValue = (value: number | null | undefined): string => {
```
bohdangarchu, line 27: move to operations or utils

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+  if (value === null || value === undefined) return 'N/A';
+  return `${value.toFixed(1)}%`;
+};
+
+const nutritionFill = (value: number | null): string => {
```
bohdangarchu, line 32: operations

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+        const nutrientLabels: { [key: string]: string } = {
+          fe_ai: 'Iron',
+          fol_ai: 'Folate',
+          va_ai: 'Vitamin A',
+          vb12_ai: 'Vitamin B12',
+          zn_ai: 'Zinc',
+          mimi_simple: 'Mean Adequacy Ratio',
+        };
+        const nutrientLabel = nutrientLabels[selectedNutrient] || 'Unknown Nutrient';
+        const tooltipContent = `
```
bohdangarchu, line 65: I think it would be nice to have consistent styling for this tooltip. It should have dark/light bg color (depending on theme), have rounded corners and slightly transparent bg. You can take a look at tooltip styles in my pr

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+          <b>${feature.properties?.Name}</b><br/>
+          Risk of inadequate Intake of ${nutrientLabel}: ${formattedNutrientValue}
+        `;
+        layer.unbindTooltip();
+        layer.bindTooltip(tooltipContent, { className: tooltip?.className || 'state-tooltip' });
+      }
+    });
+  }, [selectedNutrient]);
+
+  const dynamicStyle = (feature: GeoJSON.Feature<Geometry, GeoJsonProperties> | undefined): PathOptions => {
```
bohdangarchu, line 75: operations

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+    return {
+      fillColor: nutritionFill(value),
+      color: '#000',
+      weight: 1,
+      fillOpacity: 0.6,
+    };
+  };
+
+  const onEachFeature = (feature: GeoJsonProperties, layer: L.Layer): void => {
```
bohdangarchu, line 90: operations

src/app/Nutrition/NutritionStateLayer.tsx
bohdangarchu, line 1: call the file according to component name

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+      )}
+      <div className="absolute bottom-5 right-0 z-50 pr-10">
+        <LegendContainer items={[legendItem]} />
+      </div>
+      <GeoJSON data={regionData} style={dynamicStyle} onEachFeature={onEachFeature} />
+    </>
+  );
+}
+
+StateChoropleth.getAccordionData = function (
```
bohdangarchu, line 153: operations

bohdangarchu (changes requested):
Great job! In addition to comments I noticed two things:
- I don't think we need this square border around the country
- on the first click half of the country is not shown
image

ArmanpreetGhotra:
> Great job! In addition to comments I noticed two things:
> 
> * I don't think we need this square border around the country
> * on the first click half of the country is not shown
>   image

-> Strange, i never came across such issue, although i test it out multiple times and @plaume8  did testing too.

src/components/Buttons/CustomButton.tsx
```diff
@@ -9,12 +9,17 @@ import { CustomButtonProps } from '@/domain/props/CustomButtonProps';
...
   const { variant } = attributes;
 
   return (
     <Button
       {...attributes}
+      style={{
+        backgroundColor: color,
+        minWidth: '120px',
+        textAlign: 'center',
+      }}
```
marinovl7, line 22: This should be gone. First you should not be using style, until explicitly neccessary, but tailwind classes + clsx function . Nevertheless please do not change the Custom Button component without asking/communicating your special use case because this would change everything in the app

src/components/Map/Map.tsx
```diff
@@ -1,14 +1,62 @@
...
+  const getFillColor = (dataType: string): string => {
+    switch (dataType) {
+      case 'actual':
+        return '#FFB74D';
+      case 'predicted':
+        return '#E3F2FD';
+      default:
+        return '#52525b';
+    }
+  };
```
marinovl7, line 30: this should not be here, operations

src/components/Map/Map.tsx
```diff
@@ -1,14 +1,62 @@
...
+
+  // Fetch and preprocess data
+  useEffect(() => {
+    GlobalRepo.getNutritionData()
+      .then((global) => {
```
marinovl7, line 35: we are not using promise.then, instead use await + the data should not be fetched there if that is the use case, but in the parent component of the map, which is a server component. However this data could be fetched conditionally, if the `selectedMapType = "Nutrition"`, but again in the parent component

marinovl7, line 35: What is global -> use precise naming

src/components/Map/Map.tsx
```diff
@@ -1,14 +1,62 @@
...
+                weight: 1,
+                fillOpacity: 0.5,
+                fillColor: getFillColor(item.data_type),
+              };
+              return acc;
+            },
+            {} as { [key: number]: L.PathOptions }
+          );
+
+          setCountryStyles(styles);
```
marinovl7, line 50: Why are you doing this ?

src/components/Map/Map.tsx
```diff
@@ -25,6 +73,27 @@ export default function Map({ countries, disputedAreas }: MapProps) {
...
+            const style = countryStyles[c.properties.adm0_id];
+
+            return style ? (
+              <NutritionChoropleth
+                key={c.properties.adm0_id}
+                data={{ type: 'FeatureCollection', features: [c as Feature<Geometry, GeoJsonProperties>] }}
+                style={style}
+                handleClick={(sourceTarget) => {
+                  console.log('Country clicked:', sourceTarget);
+                }}
```
marinovl7, line 89: no need of this

src/components/Map/VectorTileLayer.tsx
```diff
@@ -23,12 +25,11 @@ export default function VectorTileLayer({ countries, disputedAreas }: MapProps)
     );
     MapOperations.setMapInteractionFunctionality(baseMap);
     MapOperations.synchronizeLeafletMapbox(baseMap, mapContainer, context);
-
     return () => {
       baseMap.remove();
       context.map.off('move');
     };
-  }, [context, theme]);
+  }, [context, theme, selectedMapType]);
```
marinovl7, line 32: nope, this shouldn't go in here, this would cause the entire map to rerender when the map type changes, this should not be the case? Why did you need it?

src/app/Nutrition/NutritionStateLayer.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  const legendItem: GradientLegendContainerItem = {
+    title: 'Risk of Inadequate Micronutrient Intake',
+    startColor: 'nutritionGradientStart',
+    endColor: 'nutritionGradientEnd',
+    startLabel: '0%',
+    endLabel: '100%',
+    middleColor: 'nutritionGradientMiddle',
+    tooltipInfo: 'Shows the inadequate ratio of nutrient intake.',
+  };
```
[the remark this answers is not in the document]
plaume8: will be done in F-53

src/app/page.tsx
```diff
@@ -12,18 +12,25 @@ export default async function Home() {
   const countryMapDataPromise = globalRepo.getMapDataForCountries();
   const disputedAreasPromise = globalRepo.getDisputedAreas();
   const ipcDataPromise = globalRepo.getIpcData();
-  const [countryMapData, disputedAreas, ipcData] = await Promise.all([
+  const nutritionDataPromise = globalRepo.getNutritionData();
+  const [countryMapData, disputedAreas, ipcData, nutritionData] = await Promise.all([
```
plaume8, line 16: @marinovl7 or @bohdangarchu could you take a look and confirm if the `app/page` is the right place for the `CountryNutrion` data fetching ? -> for me it made sense to implement it there cause all other map related basic data is loaded there as well

plaume8:
@bohdangarchu  @marinovl7 - to ensure that we can merge this issue as soon as possible, I’ve created a [new issue](https://linear.app/world-food-programme-wfp/issue/F-116/nutrition-map-use-reusable-legend) where I’ll refactor the legend to make sure @ahmedfarouk2000 reusable legend is used. By then, the changes from [this pull request](https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/49/files) should also be merged, so the general legend will know which country is selected, which is also needed for the Nutrition view.

src/app/page.tsx
bohdangarchu: looks good to me

src/components/Map/NutritionChoropleth.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const [countryStyles, setCountryStyles] = useState<{ [key: number]: L.PathOptions }>({});
+
+  // given the `CountryNutrition` data -> parse the data to map country styling
+  useEffect(() => {
+    const parsedStyles = NutritionChoroplethOperations.getCountryStyles(nutritionData);
+    setCountryStyles(parsedStyles);
+  }, [nutritionData]);
+
+  if (selectedAlert) {
+    toggleAlert(selectedAlert);
```
bohdangarchu, line 35: this should only happen on country click. Right now this disables alerts for the global view and what we want is when you click on a country and zoom in then alerts are disabled

src/components/Map/NutritionStateChoropleth.tsx
```diff
@@ -0,0 +1,84 @@
...
+          direction: 'top',
+          offset: [0, -10],
+          permanent: false,
+          sticky: true,
+        });
+      }
+    });
+  }, [selectedNutrient]);
+
+  const dynamicStyle = (feature: GeoJSON.Feature<Geometry, GeoJsonProperties> | undefined): PathOptions => {
```
bohdangarchu, line 57: NutritionStateChoroplethOperations

src/components/Map/Map.tsx
```diff
@@ -62,6 +70,22 @@ export default function Map({ countries, disputedAreas, ipcData }: MapProps) {
               toggleAlert={toggleAlert}
             />
           ))}
+      {selectedMapType === GlobalInsight.NUTRITION &&
+        countries.features &&
+        countries.features
+          .filter((countryData) => countryData.properties.interactive)
```
bohdangarchu, line 76: right now every country is clickable and zoom with region split works for every country. you can filter out only clickable countries by checking if their id is in `nutritionData`
image

bohdangarchu (changes requested):
please fix these issues, otherwise lgtm

src/app/page.tsx
Tschonti: Actually, it seems to me that we can delay fetching the nutrition data until the user selects the nutrition map. It's my first time looking at this PR, so feel free to challenge me if you think I'm wrong :D

in my understanding, the nutrition data fetched here is passed to the Map component, which passes it to `VectorTileLayer` and `NutritionChoropleth`. `VectorTileLayer` just passes it to the methods of `MapOperations`, but those methods don't use it at all. So we don't need to pass it to this component at all.

`NutritionChoropleth` is only rendered when the user selects the nutrition map, so this component can fetch the data. If you use the `useNutritionQuery` hook in the component's body, it'll automatically be fetched the first time this component loads. But in this case, we have to do something with the loading state (tho this request should be really fast). But generally, we are trying to avoid loading data before it's absolutely needed, so I recommend using this approach.

The current implementation of the IPC data required the data to be there from the beginning, that's why it was fetched on the server. But this is also being reworked.

Tschonti:
The legend stays the same even when we're looking at the region view, I assume we should be using a gradient legend there

marinovl7:
Please fix the merge conflicts, and I will merge this PR as it has been here for a long time! Create another ticket to implemented the required changes as the legend, the data fetching and etc.

ArmanpreetGhotra:
@marinovl7 main branch is merged

PR #50, opened 2024-11-20
Feature/f 46 country alerts

georgi4444:
- new alert type - Country Alerts
- new icon
- positioning algo: try to position the alerts on a circle around the country center with radius 100 km. If they lay outside of the country reduce the radius by 10 km and try to position them again. Break when a position inside of the country is found or if the country center is reached.


marinovl7 (approved):
Great Job! Not a single comment I can do, perfect execution

PR #51, opened 2024-11-21
Alert loading state

Tschonti:
When the alerts are loading, a loading spinner is shown on the button.

Also added tanstack devtools with this PR, it comes handy sometimes.


marinovl7 (approved):
LGTM!

PR #56, opened 2024-11-23
fix: replace info.svg of accordion with info icon of iconsax-react

ArmanpreetGhotra:
replace info.svg in accordions with info icon from react library


jschoedl:
This is a different icon (! instead of i) - is this what you wanted to do? I think just keeping the svg and dynamically coloring it as you do in #52 should be fine, as there's no real equivalent in the library.

marinovl7 (changes requested):
I agree with Jakob, this is not really the icon we want. Instead you can creat a custom component, here you are

```typescript
import React from 'react'
import { InfoCircle } from 'iconsax-react'

const CustomInfoCircle = (props: React.ComponentProps<typeof InfoCircle>) => {
  return (
    <InfoCircle
      {...props}
      style={{
        transform: 'rotate(180deg)',
        ...props.style,
      }}
    />
  )
}

export default CustomInfoCircle
```

Also please check the Abbreviation.tsx component in the About and Data_Sources pages, there the icon is missing because you deleted it, and it needs to be changed there too

ArmanpreetGhotra:
@jschoedl you are right, i didn't noticed that. Now it's fixed. @marinovl7 thnx for providing me the code. Requested changes are implemented.

jschoedl (approved):
I fixed the alignment on the Data Sources page and modified @marinovl7's code to use tailwind instead of classes (this was probably ChatGPT, haha). LGTM now.

marinovl7 (approved):
LGTM!

PR #57, opened 2024-11-23
Feature/f 74 legend mobile

Esoteriker:
- integrate legend content into popup modal


src/components/Legend/LegendContainer.tsx
```diff
@@ -22,5 +33,48 @@ export default function LegendContainer({ items, loading = false }: LegendContai
         }))}
       />
     </div>
+  ) : (
+    <>
+      <div className="absolute z-50 bottom-3 right-16">
+        <Tooltip text={INFO_POPUP_TOOLTIP}>
```
bohdangarchu, line 39: what is the point of a tooltip if we are on a mobile device?

src/components/Legend/LegendContainer.tsx
```diff
@@ -22,5 +33,48 @@ export default function LegendContainer({ items, loading = false }: LegendContai
...
+          <Button
+            onClick={() => setInfoPopup(true)}
+            className="
+        relative flex items-center justify-center min-w-12 h-12 px-1 rounded-full bg-content1 shadow-md"
+          >
+            <Information size={32} />
```
bohdangarchu, line 45: not sure about this icon, it indicates something urgent to me. Maybe we can use `InfoIcon.svg` here? @marinovl7 what do you think?

bohdangarchu (changes requested):
Two minor comments, otherwise good job!

src/components/Legend/LegendContainer.tsx
Esoteriker: Since in our mechanism we only know it is in mobile version depends on the resolution, when screen width smaller than 640 px e.g. If the use open the website in small screen in laptop, then we need show tooltip although is very unusual.

src/components/Legend/LegendContainer.tsx
Esoteriker: Since in our mechanism we only know it is in mobile version depends on the resolution, when screen width smaller than 640 px e.g. If the use open the website in small screen in laptop, then we need show tooltip although is very unusual.

src/components/Legend/LegendContainer.tsx
marinovl7: See my comment in #56. Lets wait on that solution!

marinovl7 (approved):
Legit! Great Implementation and solution approach! Just one smal textual adjustment needed and lets wait for the custom info icon in #56 to merge this

bohdangarchu (approved):

marinovl7:
Merged already, you can use this icon!

marinovl7 (approved):
LGTM!

PR #59, opened 2024-11-24
feature/f-91-remove-delay-from-tooltip

plaume8:
goal: 
- removing delay from tooltip -> no delay by default
- delay can be passed as a parameter


bohdangarchu (approved):
lgtm!

marinovl7 (approved):
LGTM! Thinking about generics and configuration as always, great solution approach mindset!

PR #60, opened 2024-11-24
fix: fix margins for AlertsMenu, MapLegend, Sidebar

georgi4444:
- moved alerts menu above the mapbox logo
- fixed the map legend hiding the zoom control

Mapbox logo could potentially be moved to the right too

### DIFF
--- a/src/components/Legend/MapLegend.tsx
+++ b/src/components/Legend/MapLegend.tsx
@@ -19,7 +19,7 @@ export default function MapLegend() {
   }, [selectedAlert, selectedMapType]);
 
   return (
-    <div className="absolute bottom-5 right-0 z-50 pr-10">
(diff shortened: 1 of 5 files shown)


src/components/Legend/MapLegend.tsx
```diff
@@ -19,7 +19,7 @@ export default function MapLegend() {
   }, [selectedAlert, selectedMapType]);
 
   return (
-    <div className="absolute bottom-5 right-0 z-50 pr-10">
+    <div className="absolute bottom-0 right-0 z-50 mr-10 mb-5">
```
georgi4444, line 22: Margin instead of paddng so it doesnt hide the zoom control making it unusable

marinovl7 (approved):
LGTM!

PR #61, opened 2024-11-24
custom zoom control

Tschonti:
also added a subtle hover effect to the buttons on the map (alert menu, chatbot opener and zoom)


marinovl7 (changes requested):
Add background shadow such as the sidebar, legends and accordions. Otherwise the zoom component is not visible if the zoom control is displayed over countries colored in white in light mode. The rest is lgtm!

PR #62, opened 2024-11-24
feat: init whole ipc map view

ahmedfarouk2000:
- Most of old implementation of IPC has been removed and I have switched to leaflet layers.
- Created a new context to know which countryId is selected `SelectedCountryIdContext.tsx `, this helped me to display different legends based on which view. This also will help us in the future to toggle between different map types in the country view.
- to access ipc country view in old application (To verify data are correct) just switch to any map type and select the country then switch the map type to ipc.
- I didn't dispaly Nutrition accordion since it has always no data.
- fixed icon in accordiion based on the theme. 

image

image


bohdangarchu (changes requested):
Awesome job! Cool that we now have SelectedCountryIdContext. Two things I noticed:
- alerts are not disabled in country view
- you cannot select any other country in country view and need to first zoom out to do so

marinovl7 (approved):
LGTM! Just apply the changes as discussed like for the accordion data, if no data display No data message similar as currency exchange in FCS. Country clicking when country already clicked similar as in FCS.

src/operations/map/IpcChoroplethOperations.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const updatedFeatures = mapData.features.map((feature: CountryMapData) => ({
+      ...feature,
+      properties: {
+        ...feature.properties,
+        ipcData: ipcDataById[feature.properties.adm0_id] || null,
+      },
+    }));
+
+    return { type: 'FeatureCollection', features: updatedFeatures as Feature<Geometry, GeoJsonProperties>[] };
+  };
```
Tschonti, line 62: would it be possible to filter here so that only those countries are returned where there's IPC data? It looks like a second layer is added to the non-ipc countries as well (the border gets thicker)

ahmedfarouk2000: yeah thats true actually. Nice observation 👏 👏

ahmedfarouk2000:
all comments are resolved, the only minor change is that on country select the global layer is visable, this make sense since if user wants to toggle to different countries in this view, he won't know if this country is in ipc or not. I have removed only the global layer for the selected country to avoid having more than one layer with different colors thus user will be confused. 

image

bohdangarchu:
> all comments are resolved, the only minor change is that on country select the global layer is visable, this make sense since if user wants to toggle to different countries in this view, he won't know if this country is in ipc or not. I have removed only the global layer for the selected country to avoid having more than one layer with different colors thus user will be confused.

could you actually do it so that the global layer is invisible on country select? The issue with ipc/non-ipc countries will be handled in a separate ticket[ f-89](https://linear.app/world-food-programme-wfp/issue/F-89/clickable-countries-in-global-view) and I think it would give us a cleaner look. Also the original tool disables global view as well so we might want to follow this

Tschonti:
The only thing left from my POV is that we no longer have to fetch the ipc data on the server. It's passed to the VectorTileLayer and the MapOperations, but they don't actually use it. So the IpcLayer component could be responsible for fetching the data and this way the data won't be loaded until the user actually selects the IPC map.

You can do this by using the `useIpcQuery()` hook in the component's body. But keep in mind, now we have to deal with the loading state somehow.

It might be best to do this refactor in a separate PR so this awesome feature can be merged earlier. The same thing has to be done with the nutrition map, so maybe both tasks could be done in one ticket.

ahmedfarouk2000:
> The only thing left from my POV is that we no longer have to fetch the ipc data on the server. It's passed to the VectorTileLayer and the MapOperations, but they don't actually use it. So the IpcLayer component could be responsible for fetching the data and this way the data won't be loaded until the user actually selects the IPC map.
> 
> You can do this by using the `useIpcQuery()` hook in the component's body. But keep in mind, now we have to deal with the loading state somehow.
> 
> It might be best to do this refactor in a separate PR so this awesome feature can be merged earlier. The same thing has to be done with the nutrition map, so maybe both tasks could be done in one ticket.

This is the only thing I left from the previous implementation, we can definetly do that in different PR

ahmedfarouk2000:
> > all comments are resolved, the only minor change is that on country select the global layer is visable, this make sense since if user wants to toggle to different countries in this view, he won't know if this country is in ipc or not. I have removed only the global layer for the selected country to avoid having more than one layer with different colors thus user will be confused.
> 
> could you actually do it so that the global layer is invisible on country select? The issue with ipc/non-ipc countries will be handled in a separate ticket[ f-89](https://linear.app/world-food-programme-wfp/issue/F-89/clickable-countries-in-global-view) and I think it would give us a cleaner look. Also the original tool disables global view as well so we might want to follow this

It does not make sense to me, this means I can't also display the tooltip on other countries hover which kinda annoying until I leave the country view? maybe we can discuss this with @marinovl7

bohdangarchu:
> > > all comments are resolved, the only minor change is that on country select the global layer is visable, this make sense since if user wants to toggle to different countries in this view, he won't know if this country is in ipc or not. I have removed only the global layer for the selected country to avoid having more than one layer with different colors thus user will be confused.
> > 
> > 
> > could you actually do it so that the global layer is invisible on country select? The issue with ipc/non-ipc countries will be handled in a separate ticket[ f-89](https://linear.app/world-food-programme-wfp/issue/F-89/clickable-countries-in-global-view) and I think it would give us a cleaner look. Also the original tool disables global view as well so we might want to follow this
> 
> It does not make sense to me, this means I can't also display the tooltip on other countries hover which kinda annoying until I leave the country view? maybe we can discuss this with @marinovl7

yeah you are right, we just discussed it with Lachezar, we will leave it as its currently implemented and if we receive feedback about this we will change it. And again good job on the pr!

bohdangarchu (approved):

PR #63, opened 2024-11-25
feat: adjusted tooltip background color and text color

plaume8:
I have adjusted the tooltip colors as follows:

Screenshot 2024-11-25 at 10 23 06
Screenshot 2024-11-25 at 10 23 13

---

One general thing I noticed: the colors for the chart are hardcoded in `LineChartOperations`. This ideally shouldn’t be the case, I guess. However, I only see the option to access the NextUI colors using the following approach. 

```
import tailwindConfig from '../../../tailwind.config';
const lightColors = tailwindConfig.plugins[0].themes?.light.colors;
...
```

Should I refactor this quickly so that all chart colors depend on the NextUI colors, or should I leave it as it is now?


marinovl7:
@Tschonti I guess we had similar case before. How did you solve this issue? It was something like: 

`var(--nextui-bla-bla)` right?

bohdangarchu:
I had the same issue, you can use `hsl(var(--nextui-${colorVar}))`, see GradientLegend for example

plaume8:
> I had the same issue, you can use `hsl(var(--nextui-${colorVar}))`, see GradientLegend for example

thank you @bohdangarchu -> worked perfectly

marinovl7 (approved):
LGTM!

PR #64, opened 2024-11-25
Feature/f 85 sidebar dsplay full screen on mobile + other imporvements

georgi4444:
- sidebar full screen on mobile
- reduced size of the navigation links in general
- faster loading for logo and the other images for the sidebar and alerts menu
- shadow when scrolling for the main sidebar content
- custom scrollbar for the whole app
- shadow for the alerts menu when outside of the sidebar for better visibility when on the map


Tschonti:
I think the sidebar should close when new select a map type or an alert on mobile

PR #65, opened 2024-11-25
fix: remove rectangle which appears when click/focus over the region

ArmanpreetGhotra:
I deactivated the outline in global.css which is enabled by default by assigning outline: none.


marinovl7 (approved):
LGTM!

PR #67, opened 2024-11-25
memoize AlertContainer

Tschonti:
The rerendering of the AlertContainer (well, mostly the ConflictLayer) takes a long time, so we want to avoid rerenders unless the props or the context values have changed. 

(The map type change was slow previously when the conflict alerts were selected, that happened because the Map component is subscribed to the selectedMap changes, so on every change it rerendered itself and it's children, including the alert layers.)


georgi4444:
Such a simple fix but it works perfectly! Good job!

georgi4444 (approved):

marinovl7 (approved):

PR #68, opened 2024-11-25
feat: allow multiple accordions to expand at once

ArmanpreetGhotra:
now multiple accordions can expanded at once in LegendContainer, About and DataSources


marinovl7 (approved):
LGTM!

PR #69, opened 2024-11-26
feat: search bar

bohdangarchu:
I added a generic search bar component and put it in our side bar.

### DIFF
--- a/src/app/elements/page.tsx
+++ b/src/app/elements/page.tsx
@@ -1,7 +1,12 @@
+'use client';
+
+import { useState } from 'react';
+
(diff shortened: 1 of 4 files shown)


Tschonti:
this is going to conflict with #37, but the search icon is nice

bohdangarchu:
> this is going to conflict with #37, but the search icon is nice

true, I can remove it from the side bar, the main reason to create this was for the download portal. And Ahmet can integrate the same icon in autocomplete using `startContent`

marinovl7 (approved):

PR #71, opened 2024-11-26
Feature/F-66 Wiki page & Disclaimer

jschoedl:
Added /wiki and /disclaimer.

There are now 4 really similar `layout.tsx` files. Maybe introducing [route groups](https://nextjs.org/docs/app/building-your-application/routing/route-groups#creating-multiple-root-layouts) could be a good idea for the future (one layout for the home page, one for the elements, one for the rest), although I'm not sure how this would work out with the different metadata.


marinovl7 (approved):
LGTM!

PR #82, opened 2024-11-29
Feature/f 78 country name tooltip on country hover

plaume8:
https://linear.app/world-food-programme-wfp/issue/F-78/country-name-tooltip-on-country-hover

- country hover tooltip implemented for all different map types
- for RAINFALL and VEGETATION the country name is shown for every country
- for all other types the country name is only shown if data available


src/components/Map/VectorTileLayer.tsx
```diff
@@ -27,12 +27,17 @@ export default function VectorTileLayer({ countries, disputedAreas, ipcData, nut
       { countries, disputedAreas, ipcData, selectedMapType, nutritionData },
       mapContainer
     );
-    const popover = new mapboxgl.Popup({ closeButton: false, closeOnClick: false });
+    const popover = new mapboxgl.Popup({
```
bohdangarchu, line 30: can you also make this popover sticky like in other map types?

src/operations/map/MapOperations.tsx
```diff
@@ -309,4 +321,15 @@ export class MapOperations {
     }
     baseMap.removeLayer(layerToRemove.id);
   }
+
+  /**
+   * Create a 'HTMLDivElement' rending the given 'countryName' within a 'CountryHoverPopover'.
+   * Needed cause leaflet tooltips or mapbox popups does not accept React components.
+   */
+  static creatCountryNameTooltipElement(countryName: string): HTMLDivElement {
```
bohdangarchu, line 329: creatE

src/components/Map/VectorTileLayer.tsx
```diff
@@ -27,12 +27,17 @@ export default function VectorTileLayer({ countries, disputedAreas, ipcData, nut
...
+      closeButton: false,
+      closeOnClick: false,
+      anchor: 'bottom',
```
bohdangarchu, line 33: maybe also change anchor to left/right to stay consistent with other map types

bohdangarchu (changes requested):
lgtm!

src/components/Map/VectorTileLayer.tsx
plaume8: @bohdangarchu should be fixed - please test it again on your setup

src/components/Map/VectorTileLayer.tsx
plaume8: I ve set the alignment to 'left.' Automatically switching between left and right would require quite a bit of custom code since this isn’t supported by Mapbox by default. https://docs.mapbox.com/mapbox-gl-js/api/markers/#popup

Is this okay for you? If not, I m happy to implement the "switiching", just wanted to point this out beforehand.

bohdangarchu (approved):

plaume8:
@marinovl7 merged latest main -> ready to be checked and merged

PR #83, opened 2024-11-29
feat: add country name and global view back button

ArmanpreetGhotra:
-> Add Country name above accordions. Country Name is packed inside non expandable Accordion.

-> There is also on the right side near chatbot icon "Global View " button which go back to the global view.


marinovl7 (commented):
I reworked the PR, because the button could have been extracted as one component for reusability. Also some functionality like unzooming was missing. Some styles adjustments were done as well

PR #85, opened 2024-11-29
fix: fix color of not analyzed data in nutrition view

ArmanpreetGhotra:
Change the color of "not analyzed data" in nutrition legend


bohdangarchu (approved):

marinovl7 (changes requested):
The color of not analyzed data should be blue in dark mode and white in light mode. Since the white will be invisible, you can add an option to proide like tiny black border to it

ArmanpreetGhotra:
beforehand i discussed bohdan about colors. I'll wait until lachezar you discuss color with lukas.

PR #86, opened 2024-11-30
feat: add striped pattern to disputed areas

georgi4444:
- disputed countries borders are dashed
- where they share border with non disputed country the solid border from the country overlaps with dashed border. Tried different solutions but didn't work out. The best solution for now is to just display striped pattern additionally so it is clear that it is a disputed area


Tschonti:
I like the striped pattern, great solution. 
FYI, we will probably try to remove mapbox from our application and implement layers like these in leaflet as well. There will be a separate task for this conversion, but if you have the time you could try to implement this layer with leaflet

PR #87, opened 2024-11-30
Feature/f 111 integrate global country search with current app

Lukas0912:
- fetching the data for the regions is now managed centrally which also allows reducing code duplication
- the searchbar is now zooming to the selected country and automatically updating according to the selectedCountryId
- the interaction functionality is now reduced more towards leaflet (mapbox hovering and active/inactive countries disabled)


Tschonti:
Nice work, good solution considering all the difficulties.
I think the options from the leaflet map that improved the touchpad zooming are missing from this branch.

PR #94, opened 2024-12-02
feat: added line chart y-axis labels

plaume8:
Checked all line charts again - did the following changes:
- Added y-axis labels:
    - Headline and food inflation: “Rate in %”
    - Balance of Trade: “Mill”
    - Currency Exchange: “Exchange rate”
- Changed Balance of Trade number formatting to “millions”:

open: @marinovl7 @bohdangarchu - I noticed that on the original HungerMap website, all the line charts are 'rounded.' I’ve already implemented the option to enable this rounding in our line chart component. I could activate it for all charts? However, theoretically, a non-rounded line chart is, of course, ' statistically cleaner’/‘more scientific’ ?


marinovl7 (approved):
Top! I agree with you, our current solution is better and more "correct"

PR #98, opened 2024-12-02
feat: AccordionContainer max width

plaume8:
https://linear.app/world-food-programme-wfp/issue/F-136/accordion-should-take-the-width-of-the-parent-element


marinovl7 (approved):
LGTM! Just added something to make the tables better by enlarging them to use parent-width when the screens are bigger and have scroll effect on smaller screens

marinovl7 (approved):

PR #99, opened 2024-12-02
fix: legend accordions not opening properly

Lukas0912:
- fix: when changing alert opening the legend accordion was not working properly


marinovl7 (approved):
LGTM! Good catch!

PR #103, opened 2024-12-02
Feature/f-122 Search in Docs

jschoedl:
Search for /about, /data_sources and /wiki.


jschoedl:
Left to do:
- [ ] search on Data Sources page
- [ ] search on Wiki page
- [ ] show "no results" if there are no results
- [ ] make the search bar stick to the top while being used

further remarks:
* x-Button of the SearchBar does not work
* one middle line in AccuracyTable appears to be missing
* the analytics dialog should not move when scrolling

jschoedl:
I could use this branch's deployed version in case this doesn't get merged before the demo. I'm so sorry for the further delays, this is really taking longer than I expected.

Tschonti (commented):
Impressive work! Just a few minor bugs I found:

- the clear button in the search bar doesn't work
- the tables are empty in the data sources page when the search is empty (including the initial load)
- a new event is pushed to the browser's history for every keystroke you type in the search, which makes using the back/forward buttons of the browser really tedious. I'd suggest only pushing to the history after a debounce.

Something else to consider: these pages are now client components, so we lose SSR. This searching is a really powerful feature, so it's worth it, but I think it'd make sense to try to implement this by keeping the pages as server components. The whole content could be rendered on the server assuming there was no search and once it loads on the client, react hydration could take over to handle the searching. This can be a separate task, I'm not even sure it's possible.

Tschonti:
> the analytics dialog should not move when scrolling

I created an issue for this, I'll handle it

jschoedl:
@Tschonti Thanks, I fixed the points you commented on.

Regarding SSR / 'use client': I am currently inserting a `Highlighter` component around every text that can be highlighted. Those inserted components need to get the search text - currently that happens using the query param. As a result, most of the page content is interactive and thus needs to be client-based (even if we managed to make the `page.tsx` files server components).

I can think of two alternatives:
* We could render the raw content without the highlight ability and replace it with highlightable text once the user types something into the search bar.
* If we still plan to do a global search feature, we could similarly remove the highlight ability, but use a different component for the search results incl. highlighting.

I'm not 100% sure though if there aren't other places where this wouldn't work.

Tschonti:
the global search feature is no longer planned.
Honestly, I don't think it's worth that much effort to make the SSR work, even though the customer got really excited about it at the meeting :D 
These are not the most important pages of the app and they won't be that slow with client-side rendering either

src/app/about/page.tsx
```diff
@@ -1,47 +1,56 @@
...
+            key={key}
+            heading={heading}
+            textElements={textElements}
+            accordionItems={accordionItems}
+            searchWords={searchWords}
+            onVisibilityChange={(visible: boolean) =>
+              setVisibleAccordions((prevState) => {
+                if (visible) prevState.add(key);
+                else prevState.delete(key);
+                return prevState;
```
marinovl7, line 46: extract this into function

src/app/about/page.tsx
```diff
@@ -1,47 +1,56 @@
...
+
+  const [searchWords, setSearchWords] = useState<string[]>([]);
+  const [visibleAccordions, setVisibleAccordions] = useState<Set<string>>(new Set());
+
   return (
-    <>
-      <section>
-        <h1>
-          About <HungerMapLiveSuperscript />
+    <Suspense>
```
marinovl7, line 25: Add PageSkeleton to the Suspense, currently nothing is shown, not even indicating that we are loading data. Please add Skeleton for the searchBar, some text paragraphs and accordions. Maybe take inspiration from Linkedin or Notion Loading indicators there

src/app/data_sources/page.tsx
```diff
@@ -1,28 +1,28 @@
...
-        Additional sources used as input variables for the predictive model but not for display purposes are listed on
-        the{' '}
-        <StyledLink href="/about" isInternal>
-          About page
-        </StyledLink>
-        .
-      </p>
-      <AccordionContainer items={dataSourceAccordionItems} multipleSelectionMode />
-    </section>
+    <Suspense>
```
marinovl7, line 15: Skeleton

src/app/wiki/page.tsx
```diff
@@ -1,24 +1,28 @@
...
-        This page contains more detailed explanations for some of the concepts behind the <HungerMapLiveSuperscript />.
-        See{' '}
-        <StyledLink href="/about" isInternal>
-          About
-        </StyledLink>{' '}
-        for a broader overview.
-      </p>
-      <AccordionContainer items={wikiEntries} />
-    </section>
+    <Suspense>
```
marinovl7, line 15: Skeleton

src/components/Search/DocsSearchBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  useEffect(() => {
+    router.push(`${pathname}?search=${debouncedSearch}`);
+  }, [debouncedSearch]);
+
+  return (
+    <div className="h-10">
+      <SearchBar
+        value={search}
+        onValueChange={(v) => {
```
marinovl7, line 25: precise naming, what is v?

src/components/Search/RecursiveHighlighter.tsx
marinovl7, line 1: Impressive!

src/domain/constant/about/aboutTextElements.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import LiveSuperscript from '@/components/About/LiveSuperscript';
+import StyledLink from '@/components/About/StyledLink';
+import { SearchOperations } from '@/operations/Search/SearchOperations';
+
+export const aboutTextElements = SearchOperations.makeTextElementsSearchable([
```
marinovl7, line 7: do same as other exported default elements to stay consistent and its much easier for non coders
```typescript
const aboutTextElements = ...
...
...
export default SearchOperations.makeAccordionItemsSearchable(aboutTextElements);

File: src/domain/constant/wiki/wikiTextElements.tsx [hunk 7]
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import LiveSuperscript from '@/components/About/LiveSuperscript';
+import StyledLink from '@/components/About/StyledLink';
+import { SearchOperations } from '@/operations/Search/SearchOperations';
+
+export const wikiTextElements = SearchOperations.makeTextElementsSearchable([
```
marinovl7, line 7: same here, consistency

marinovl7 (changes requested):
Crazy Feature! Really, amazing! Just few small comments, and could I please ask you to add docs how the search words, how we could change the words search from contains to equals or etc, or how for example to introduce more pages with the same Search Feature. I think you will forget after while and makes sense to do this right now, as I think this will be used further for sure!

marinovl7 (approved):

PR #117, opened 2024-12-08
Feature/f 146 integrate chatbot reports chatting

ahmtslmngcl:
# Implement Report Chat Functionality and According Chatbot Accessibility 

# Changes Introduced:
### Chat with Reports in DownloadPortal
- Added functionality to chat with report contents when the button in the DownloadPortal is clicked.
- Implemented PDF text extraction to retrieve information from reports, which is sent to the backend (BE) for further processing.
- Updated the BE system prompt to ensure the chatbot understands it's interacting with report data (refer to corresponding BE Repo PR for details).

### Global Chatbot Access
- Integrated the chatbot into the top bar, making it accessible across all pages of the application.
- Implemented local data storage of the chats for 1 week which also ensures conversation consistency when navigating between sites.

---
I also explored alternative methods for extracting additional PDF information, including OCR and OpenAI Assistant, text extraction approach seems to be effective and sufficient for our case

Edit: Chatbot was removed from the Topbar component itself and was placed into the layout files of the pages. The visual positioning still is the same. This allows easier usage of the Topbar without the Chatbot for example in the error page later on.


src/components/Chatbot/Chatbot.tsx
```diff
@@ -302,6 +307,11 @@ export default function HungerMapChatbot() {
                             ? 'rounded-xl p-2 bg-chatbotUserMsg dark:bg-chatbotUserMsg ml-12'
                             : 'bg-transparent pl-2 pr-2'
                         )}
+                        style={{
+                          wordWrap: 'break-word',
+                          overflowWrap: 'break-word',
+                          whiteSpace: 'normal',
+                        }}
```
marinovl7, line 314: use className from tailwind, we do not use the style prop

src/components/DownloadPortal/CountryReports.tsx
```diff
@@ -33,6 +34,7 @@ export default function CountryReports({ countryCodesData }: CountryReportsProps
         data={DownloadPortalOperations.formatTableData(
           filteredData,
           setSelectedCountry,
+          useChatbot().chatWithReport,
```
marinovl7, line 37: Extract this into const

```typescript
const chatBot = useChatbot()
const { chatWithReport } = chatbot
```

src/domain/contexts/ChatbotContext.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { useMediaQuery } from '@/utils/resolution';
+
+interface ChatbotContextType {
+  chats: IChat[];
+  setChats: React.Dispatch<React.SetStateAction<IChat[]>>;
+  currentChatIndex: number;
+  setCurrentChatIndex: React.Dispatch<React.SetStateAction<number>>;
+  openChat: (chatIndex: number) => void;
+  startNewChat: (newChat?: IChat) => void;
+  chatWithReport: (countryName: string, report: string) => Promise<void>;
```
marinovl7, line 15: chatWithReport is a little bit weird naming. Can it be more precise? At least I don't understand what it is about

src/operations/chatbot/Chatbot.ts
```diff
@@ -15,4 +15,71 @@ export default class ChatbotOperations {
...
+      // If no stored data, return default chat
+      return [
+        {
+          id: 1,
+          title: 'Chat 1',
+          messages: [],
+          isTyping: false,
+          timestamp: Date.now(),
+        },
+      ];
```
marinovl7, line 61: code duplication with the thing above

src/operations/chatbot/Chatbot.ts
```diff
@@ -15,4 +15,71 @@ export default class ChatbotOperations {
...
+
+    return [
+      {
+        id: 1,
+        title: 'Chat 1',
+        messages: [],
+        isTyping: false,
+        timestamp: Date.now(),
+      },
+    ];
```
marinovl7, line 72: code duplication with the thing above

src/operations/chatbot/Chatbot.ts
```diff
@@ -15,4 +15,71 @@ export default class ChatbotOperations {
...
+  static saveChatsToStorage(chats: IChat[]): void {
+    if (typeof window !== 'undefined') {
+      const chatsWithTimestamp = chats.map((chat) => ({
+        ...chat,
+        timestamp: chat.timestamp || Date.now(),
+      }));
+
+      localStorage.setItem('chatbotChatsWithTimestamp', JSON.stringify(chatsWithTimestamp));
+    }
+  }
```
marinovl7, line 84: Maybe add error handling if window is undefined with the Snackbar

src/operations/chatbot/ExtractPdfText.ts
```diff
@@ -0,0 +1,44 @@
...
+      );
+    }
+
+    const pageTexts = await Promise.all(pagePromises);
+    fullText = pageTexts.join('\n');
+
+    return fullText.trim();
+  } catch (error) {
+    throw new Error(`Error extracting text from PDF: ${error}`);
+  }
```
marinovl7, line 43: Better Error Handling with the Snackbar is required

marinovl7 (changes requested):
Overall valuable improvements, thanks for thinking about the context and local storage! Also one more thing I noticed, on mobile the chatbot icon in the Topbar is displayed over the theme toggle, I think it would be nice to actually put the theme toggle in the mobile bar for smaller screens. Could you do that as well?

PR #118, opened 2024-12-08
feat: implement loading skeleton for map legend component

georgi4444:


Tschonti (approved):
Nice job!

The only improvement I could think of is the color, I think the skeleton is using content2, while we're using content1 as the background. This is also the case for the HungerAlert skeleton, so if you could fix those 2, that'd be great

PR #119, opened 2024-12-08
feat: add loading to accordions of download portal with the help load…

ArmanpreetGhotra:
loading indicator to accordion items of download portal with the help of loading.tsx.
If network is 3g, originally you can't open accordion. 
This is what it looks like right now
Screenshot 2024-12-08 at 21 53 47

### DIFF
--- a/src/app/download-portal/loading.tsx
+++ b/src/app/download-portal/loading.tsx
@@ -1,0 +1,23 @@
+import AccordionContainer from '@/components/Accordions/AccordionContainer';
+import { TITLE } from '@/domain/entities/download/Country';
+
+export default function Loading() {
+  const loading = true;
+
+  return (
+    <div>
+      <h1>Download Portal</h1>
+      <AccordionContainer
+        items={[
+          {
+            title: 'Country Reports',
+          },
+          {
+            title: TITLE,
+          },
+        ]}
+        loading={loading}
+      />
+    </div>
+  );
+}


bohdangarchu:
I don't understand why I see it locally but don't see it in the deployed version

ArmanpreetGhotra:
that's weird, now i'm also not seeing it in the deployed version but locally i can see the loading indicator

bohdangarchu:
Maybe adm0data is cached in production while locally it doesn't get cached. I did a prod build with `cache: 'no-store'` and then it worked. @Tschonti do you have an idea why we are not seeing loading.tsx?

Tschonti:
Correct, my current theory is that in prod environment the adm0data is cached. Thanks to this question, I found [this](https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/125), so soon it will be cached in dev as well. But it's still a good idea to create loading pages, since some users will see it when the cache is cold

bohdangarchu:
> Correct, my current theory is that in prod environment the adm0data is cached. Thanks to this question, I found [this](https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/125), so soon it will be cached in dev as well. But it's still a good idea to create loading pages, since some users will see it when the cache is cold

Great, thanks!

bohdangarchu (approved):

PR #121, opened 2024-12-10
Feature/f 155 charts download button should be available on minimized

plaume8:
[F-155](https://linear.app/world-food-programme-wfp/issue/F-155/charts-download-button-should-be-available-on-minimized-chart-as-well)

I ve added the download button to the "small" mode. Additionally, I introduced the `disableDownload` parameter to the LineChart component, allowing the download button to be disabled.

One issue came to my attention that I hadnt noticed before: using colors with `hsl(var(--nextui-...))` causes problems with Highcharts downloads. Highcharts cannot process colors in the HSL format for exporting, which results in exported PNG or SVG charts missing colors -> so some are rendered in black and white, and sometimes entire series are missing.
I tried to find a solution but didnt come across anything except the option to convert HSL using a custom method or to define colors statically using hex codes. I decided to impelment the latter, as its the simplest approach, requires no additional code, and colors only need to be defined in `LineChartOperations.getHighChartOptions()` without affecting any other part of the application.


bohdangarchu:
Great that you noticed this. I know you already implemented the second approach but do you think we can still try the first one? Basically using 
```
function getCSSVariableValue(variableName: string): string {
      return getComputedStyle(document.documentElement).getPropertyValue(variableName).trim();
    }
    console.log(getCSSVariableValue('--nextui-clusterOrange'));
```
and copying a hsl implementation from the internet

marinovl7 (approved):

plaume8:
> Great that you noticed this. I know you already implemented the second approach but do you think we can still try the first one? Basically using
> 
> ```
> function getCSSVariableValue(variableName: string): string {
>       return getComputedStyle(document.documentElement).getPropertyValue(variableName).trim();
>     }
>     console.log(getCSSVariableValue('--nextui-clusterOrange'));
> ```
> 
> and copying a hsl implementation from the internet

done - I wanted to add a new utility function and discovered, that there is already a utility function that solves the problem `tailwind-util.getTailwindColor` -> used this one and it works

bohdangarchu (approved):
lgtm!

PR #122, opened 2024-12-10
feat: add zoom control skeleton

georgi4444:


marinovl7 (changes requested):
The z-Index is too big and its displayed over the cookie consent, this should not be the case

georgi4444:
> The z-Index is too big and its displayed over the cookie consent, this should not be the case

Fixed it. I decided to copy the structure of the Map.tsx and use it for the skeleton as well. The map container has z-index 1

georgi4444:
> The z-Index is too big and its displayed over the cookie consent, this should not be the case

Fixed it. I decided to copy the structure of the Map.tsx and use it for the skeleton. The map container has z-index 1 one thats why I wrapped the zoom skeleton and the map skeleton in a div with z-index 1

Tschonti (approved):

PR #125, opened 2024-12-10
Disable cache max size

Tschonti:
source: https://stackoverflow.com/a/79229889/10719279


Tschonti:
The warning in development is gone, but I think the loading time is the same, so my theory (that the 2MB limit is not a thing in production) seems to be correct

I think it's still worth merging, tho not as important now. There also won't be any use for the two new adm0 endpoints.

marinovl7 (approved):

PR #127, opened 2024-12-11
feat: add switching logic when map type get changed

ArmanpreetGhotra:
if the country data is available, the map would stay in the same position, if not then it will back in global view. I'm checking in the mapOperations.tsx, if the data of the country for specific type of map is available or not. All the 3 maps has slightly different approach to check data.


src/components/Map/Map.tsx
```diff
@@ -82,13 +83,34 @@ export default function Map({ countries, disputedAreas, fcsData, alertData }: Ma
...
+          mapRef.current?.fitBounds(L.geoJSON(selectedCountryData as GeoJSON).getBounds(), { animate: true });
+        } else {
+          mapRef.current?.fitBounds(
+            [
+              [-90, -180],
+              [90, 180],
+            ],
+            { animate: true }
+          );
+          setSelectedCountryId(null);
```
Tschonti, line 100: instead of this, I think it's best to just to the same thing the back to global button does (zoomOut 4 I think)

src/components/Map/Map.tsx
```diff
@@ -41,6 +41,7 @@ export default function Map({ countries, disputedAreas, fcsData, alertData }: Ma
   const [regionNutritionData, setRegionNutritionData] = useState<FeatureCollection | undefined>();
   const [ipcRegionData, setIpcRegionData] = useState<FeatureCollection<Geometry, GeoJsonProperties> | undefined>();
   const [selectedCountryName, setSelectedCountryName] = useState<string | undefined>(undefined);
+  const [data, setData] = useState<boolean>(true);
```
Tschonti, line 44: rename this to something more meaningful, this doesn't actually contain any real data

Tschonti (changes requested):
Generally good direction for sure. The switch seems kinda slow, but I don't see any unnecessary network calls or anything, so I guess we can't do anything about it.

I found one major bug tho: If you click on a country that doesn't have ipc data, then switch to ipc you get put on the global map. If you now click on a country, the regions get loaded but you're not zoomed to the country.  And this happens with the other maps too.

src/components/Map/Map.tsx
ArmanpreetGhotra: i tried to do that, but i got error like useMape() can't be use here, it should be used be descendant of mapRef.

Tschonti: `mapRef.current?.zoomOut` should work

ArmanpreetGhotra:
> https://deploy-preview-127--wfp-hungermap.netlify.app

it's fixed now, the pr was in draft/progress, because i was having issue with double click and bohdan said like i can draft pr so that he can have look on that. But now all issues are fixed @Tschonti

Tschonti (approved):
LGTM

maybe we could also zoom out on the vegetation and rainfall map, not sure

ArmanpreetGhotra:
@Tschonti maybe we could add "Global View" button when the map type is vegetation or rainfall ?

bohdangarchu (approved):
lgtm!

bohdangarchu:
> @Tschonti maybe we could add "Global View" button when the map type is vegetation or rainfall ?

its already there afaik

src/components/Map/Map.tsx
```diff
@@ -82,14 +87,31 @@ export default function Map({ countries, disputedAreas, fcsData, alertData }: Ma
...
           setCountryIso3Data,
           setRegionNutritionData,
-          setIpcRegionData
-        );
-        setSelectedCountryName(selectedCountryData.properties.adm0_name);
+          setIpcRegionData,
+          setIsDataAvailable
+        ).then(() => {
+          setIsLoading(false);
+        });
```
marinovl7, line 94: We don't use then(), instead await the fetchCountryData promise and after that setIsLoading to false

src/operations/map/MapOperations.tsx
```diff
@@ -53,20 +49,43 @@ export class MapOperations {
 
       if (selectedMapType === GlobalInsight.FOOD) {
         const newCountryIso3Data = await countryRepository.getCountryIso3Data(selectedCountryData.properties.iso3);
-        setCountryIso3Data(newCountryIso3Data);
+        if (Array.isArray(newCountryIso3Data) && newCountryIso3Data[1] === 404) {
+          setIsDataAvailable(false);
```
marinovl7, line 53: This is wrong. Even if we don't have iso3 data but have adm1data we have data and shoudln't zoom out

src/operations/map/MapOperations.tsx
```diff
@@ -53,20 +49,43 @@ export class MapOperations {
...
+        const newRegionData = await countryRepository.getRegionData(selectedCountryData.properties.adm0_id);
+        const hasFcs = newRegionData.features.some((feature) => feature.properties?.fcs !== undefined);
+        setIsDataAvailable(hasFcs);
+        if (newRegionData && newRegionData.features) {
+          setRegionData({
+            type: 'FeatureCollection',
+            features: newRegionData.features as GeoJsonFeature<Geometry, GeoJsonProperties>[],
+          });
+        }
+      }
```
marinovl7, line 88: Why did you move this to the end? The adm1data should be the first request which should be triggered when clicked on a country. It is also like this on the current new HungerMap. Please put it back to the top and also consider only this endpoint for the setIsDataAvailable. Iso3 and countryData is not relevant, just the adm1data

marinovl7 (changes requested):
Overall good job! However some things are wrong and need to be redone. The adm1data request sending is super important to be the first one, as this slows the performance if it's not the first one

src/operations/map/MapOperations.tsx
ArmanpreetGhotra: i moved to the end because even if we don't have the data of country in fcs view, i always get data for the region and which make setIsDataAvailable(true). i need to move it to the end. for example ethopia, ethopia is in nutrition and ipc view but not in fcs view. but if the loop stays at the top and you make switch from ipc/nutrition view you always get true. Which shouldn't be the case

ArmanpreetGhotra:
> > @Tschonti maybe we could add "Global View" button when the map type is vegetation or rainfall ?
> 
> its already there afaik

@bohdangarchu it's not in vegetation and in rainfall

bohdangarchu:
> > > @Tschonti maybe we could add "Global View" button when the map type is vegetation or rainfall ?
> > 
> > 
> > its already there afaik
> 
> @bohdangarchu it's not in vegetation and in rainfall

image

ArmanpreetGhotra:
> > > > @Tschonti maybe we could add "Global View" button when the map type is vegetation or rainfall ?
> > > 
> > > 
> > > its already there afaik
> > 
> > 
> > @bohdangarchu it's not in vegetation and in rainfall
> 
> image

it's the current version

bohdangarchu:
> > > > > @Tschonti maybe we could add "Global View" button when the map type is vegetation or rainfall ?
> > > > 
> > > > 
> > > > its already there afaik
> > > 
> > > 
> > > @bohdangarchu it's not in vegetation and in rainfall
> > 
> > 
> > image
> 
> it's the current version [ Screenshot 2024-12-16 at 09 05 17 ](url)

yes bc you are in global view. but when you switch to veg or rainfall in country view then the button is there

marinovl7 (approved):

PR #131, opened 2024-12-13
Lighthouse best practices & accessibility

Tschonti:
Things changed:

- generate source maps in production: it's an open source app, so why not. The check still fails on the preview, I guess the env is not production, should work when merged
- fixed a few warnings (highcharts accessibility module included, deprecated meta tag removed)
- CSP: only allow scripts, styles from certain domains. This is a bit hard, because next adds a bunch of scripts and styles dynamically, so inline scripts and styles are also allowed, which kinda defeats the purpose, but Lighthouse is satisfied. Also the netlify preview features no longer work because netlify's script is not allowed, but we were'nt using that anyway

remaining issues:
-  some warnings because preloaded images or fonts are not used instantly - I think we can ignore these, we have a reason to preload them


src/components/Accordions/AccordionBoxItems.tsx
```diff
@@ -72,9 +72,23 @@ export default function AccordionBoxItems({
...
                         </Button>
+                        //   <div
+                        //   role="button"
+                        //   tabIndex={0}
+                        //   onClick={() => 1}
+                        //   onKeyDown={() => 1}
+                        //   className="w-[37px] h-[37px] p-[5.5px] hover:bg-default hover:opacity-40 rounded-xl"
+                        // >
+                        //   <span className="hover:opacity-100 hover:text-white">{item.infoIcon}</span>
+                        // </div>
```
marinovl7, line 91: Remove this please

marinovl7 (changes requested):
Just one small change. The Lighthouse score fails for accessibility, idk if its hapenning on prod, but lets have a look tomorrow in the demos and then if it works fine and we defined the next actions we will merge this

Tschonti:
best practices score should be 100 on master

PR #133, opened 2024-12-14
fix: adjust sidebar height

georgi4444:


Tschonti:
unfortunately, the alerts are still hidden on my 15.6" laptop. I think those buttons are way more important than disclaimer etc. Maybe we should collapse the links on the bottom? Or make the whole sidebar scrollable?
image

marinovl7 (changes requested):
Agree with Samuel, I would also make the whole sidebar scrollable

Tschonti:
Much better, but can we still make the whole sidebar scrollable? I think with this it looks good on most screens, but just in case someone looks at it on a really small screen, I think we should still hide the links before the alerts

georgi4444:
> Much better, but can we still make the whole sidebar scrollable? I think with this it looks good on most screens, but just in case someone looks at it on a really small screen, I think we should still hide the links before the alerts

Done this as well now

PR #136, opened 2024-12-14
Google Analytics improvements

Tschonti:
- send event when the selected country changes (it takes some time (~one day) for these event to reach GA)
- created explorations that show only our custom events
image
image
image
image

You can see them here: https://analytics.google.com/analytics/web/?authuser=3#/analysis/p468690329

Sign in with the jstwfp@gmail.com account (pwd in notion)

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -84,6 +84,9 @@ export default function Map({ countries, disputedAreas, fcsData, alertData }: Ma
           setRegionNutritionData,
           setIpcRegionData
         );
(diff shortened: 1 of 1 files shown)


Tschonti:
country alerts and fcs map have lower numbers because they're the default and no event is sent then. Should we send an event at the start just to make the "race" more fair? @marinovl7

marinovl7:
> country alerts and fcs map have lower numbers because they're the default and no event is sent then. Should we send an event at the start just to make the "race" more fair? @marinovl7

Would love to actually have a demo of your work here! We could discuss this further then

marinovl7 (approved):

PR #140, opened 2024-12-16
Feature/f 184 chart no data visualization

plaume8:
[F-184](https://linear.app/world-food-programme-wfp/issue/F-184/chart-no-data-visualization)

I also added a "chart with no available data" example to `/elements`


plaume8:
accidentally created by mistake.

PR #142, opened 2024-12-16
Feature/f 152 regions labeling fcs ipc nutrition

Lukas0912:
- the labels are loaded because they contain data about the positioning of the tooltips, calculating the position from the geojson data was not giving the same results; once the data is loaded though it won't be loaded again
- positioning dependent of the feature layer didn't work for some reason so I had to bind the tooltips to the map, which has the disadvantage that listeners and tooltips are not removed automatically; I took care of removing the visible tooltips as well as the listeners for not having any hidden performance disadvantages; -> have a look at it again though, especially regarding performance
- the labels are not perfectly dependent on the region borders since it's very hard to consider the placing of the label and the exact shape of the regions; instead the length is estimated based on the number of letters and also the sensitivity, i.e. how wide one letter is considered, can be changed


Tschonti:
The adm1_labels.geojson request again introduces the CORS issue, we have to let the customer know that they'll have to update cdn.hungermapdata.org

PR #149, opened 2024-12-19
Feature/f 189 add skeletons for fcs nutrition and ipc accordions

Lukas0912:
- I decided for the loading element of the accordions instead of skeletons since it fits to the loading style when switching alerts and codewise is the cleaner variant


marinovl7:
Merge Conflicts

Lukas0912:
Now it looks a bit weird when the data is still loading but the accordions are opening automatically. I will leave the accordions closed until the loading process is finished and then open them.

PR #154, opened 2024-12-21
Feature/f-169 Fix Hydration Errors

jschoedl:
**Initial Problem:** In some (independent) cases, the Data Source page was shown in a corrupted way and without interactivity. A reload didn't fix it. [Screenshot](https://uploads.linear.app/620c9493-591e-4baa-8cb8-dabda150c013/ad2ab5d9-b998-47fb-86c4-514c0f3bdbf8/c8a9b20f-5e1f-4f32-9824-2ed43025fdab)

**Repro:** I couldn't reproduce the issue except for one time where I had collected no details. However, the screenshot looks similar to the initial response from the server (i.e. the server-rendered initial portion of the page): [Screenshot from Devtools](https://github.com/user-attachments/assets/d105f374-f57f-4fb9-b9fc-9138220f273c)

**Fix attempt:** It seems like something went wrong right after the browser received the initial response. At this point, React hydration takes place, so I suspect that a [hydration error](https://nextjs.org/docs/messages/react-hydration-error) could have occurred. I used the [W3C HTML validator](https://validator.w3.org/) to search for invalid nested elements in the output HTML and changed the element types wherever possible. One `div` as child of a `button` remains, however this seems to be internally inside of the NextUI `AccordionItem` component.

Since I cannot reproduce the issue, I obviously cannot say for sure if this really fixes it.

### DIFF
--- a/src/components/Accordions/AccordionBoxItems.tsx
+++ b/src/components/Accordions/AccordionBoxItems.tsx
@@ -58,11 +58,11 @@ export default function AccordionBoxItems({
             className={`last:border-b-0 ${color} white:bg-white overflow-hidden shadow-md`}
             hideIndicator={noSelectionMode}
             title={
-              <div className="flex justify-between items-center w-full">
-                <div className="flex gap-4">
+              <span className="flex justify-between items-center w-full">
+                <span className="flex gap-4">
                   <Highlighter searchWords={highlightedTitleWords} textToHighlight={item.title} autoEscape />
                   {loading && <Spinner size="sm" />}
-                </div>
+                </span>
                 {item.tooltipInfo && (
                   <Tooltip text={item.tooltipInfo}>
                     {item.infoIcon && <span className="w-[37px] h-[37px] p-[5.5px]">{item.infoIcon}</span>}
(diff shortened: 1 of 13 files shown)


src/components/Accordions/AccordionBoxItems.tsx
```diff
@@ -72,7 +72,7 @@ export default function AccordionBoxItems({
                   <Popover>
                     <PopoverTrigger>
                       {item.infoIcon && (
-                        <Button isIconOnly className="w-[37px] h-[37px] p-[5.5px]" variant="light">
+                        <Button as="span" isIconOnly className="w-[37px] h-[37px] p-[5.5px]" variant="light">
```
Tschonti, line 75: wow this is nice. I spent a bunch of time trying to solve a hydration issue on the map page (we can see the warning in the console, but there were no noticeable effects), but this solves that as well. Could you add it to this PR?
`AccordionBoxItems.tsx` line 75

jschoedl: I'm on the go rn, but you could cherry-pick commit `3cd34f32c898e406003a4f7817bd22ef86d93ec6` (fix: replace nested button with span). Or otherwise it's also fine for me if you add these 10 characters yourself, lol.

PR #155, opened 2024-12-21
Feature/f 183 categorical chart minor improvements

plaume8:
- all solved and implemented minor todos are described in [F-183](https://linear.app/world-food-programme-wfp/issue/F-183/categorical-chart-minor-improvements)
- please especially check/review the selected static and increased height of the line charts in the "Comparison Portal".


marinovl7:
Merge Conflicts

bohdangarchu (changes requested):
When selecting a country on comparison portal charts are rerendered multiple times, while on develop it only happens once. Could you check this please?

plaume8:
> When selecting a country on comparison portal charts are rerendered multiple times, while on develop it only happens once. Could you check this please?

could not find a quick fix -> creates a new issue [f-223](https://linear.app/world-food-programme-wfp/issue/F-223/bug-chart-multiple-rerendering) -> such that this issue can be merged as soon as possible

src/components/Charts/helpers/ChartContainer.tsx
```diff
@@ -19,35 +20,58 @@ import ChartContainerProps from '@/domain/props/ChartContainerProps';
...
   // full screen modal state handling
   const { isOpen, onOpen, onClose, onOpenChange } = useDisclosure();
 
+  // handling the theme switch
+  useEffect(() => {
+    // `theme` change does not guarantee that the NextUI CSS colors have already been changed;
+    // therefore we synchronize the update with the next repaint cycle, ensuring the CSS variables are updated
+    const rafId = requestAnimationFrame(() => {
+      recalculateChartOptions();
+      setChartKey((prev) => prev + 1); // forces chart to remount -> chart animation will be re-triggered
```
bohdangarchu, line 64: I also checked other graphs and noticed that they are now rendered twice every time. I feel like this approach of resetting chart key is a bit problematic bc it triggers additional animation. 

I just tested this out and if you remove this line from both useEffects everything seems to work fine (incl. theme switch).

bohdangarchu (changes requested):

bohdangarchu:
> > When selecting a country on comparison portal charts are rerendered multiple times, while on develop it only happens once. Could you check this please?
> 
> could not find a quick fix -> creates a new issue [f-223](https://linear.app/world-food-programme-wfp/issue/F-223/bug-chart-multiple-rerendering) -> such that this issue can be merged as soon as possible

This issue is introduced in this pr so in my opinion it is better to fix it here

src/components/Charts/helpers/ChartContainer.tsx
plaume8: you are absolutely right, I totally forget about this change -> I fixed it -> now only the "model type switch" is using this extra setChartKey functionallity

bohdangarchu (approved):

PR #163, opened 2025-01-09
Improve panning animation

Tschonti:
Continuation of #160 

It turns out that GeoJSONs not on the default Pane are unaffected by the padding value of the SVG renderer. Previously, Panes were used to set the zIndex of each layer, because Leaflet has some weird default values. Now that we can't use Panes with the GeoJson layers, their default zIndex is 400, so I adjusted the other layers based on this.

The result is still not ideal, but I think this is the best we can do with leaflet. If you pan a lot without letting go of the mouse, you'll still see countries that don't load, but I think it's almost unnoticeable with regular use.

Keep in mind that this solution is supposedly harder to run in browsers, but I didn't notice any performance degradation. If somebody does on their machine, please let me know.


Tschonti:
I think zooming in is a bit laggy :(

PR #181, opened 2025-01-14
mocked admin permision implementation

Tschonti:
add the `NEXT_PUBLIC_FORECASTS_ENABLED=true` env var to allow forecasts and dummy authentication. This also shows a switch in the sidebar to turn on admin mode. 
In admin mode, mock forecast FCS and RCSI data are shown in the FCS accordions. Subscription topics are also filtered based on roles, though I couldn't test this.

Something to note: once authentication is properly implemented, it also has to be done on the email service backend. This solution only prevents the topics with admin role from being shown to regular users, but they can see the API response and subscribe by manually sending a POST request.


src/operations/map/FcsAccordionOperations.tsx
```diff
@@ -2,57 +2,94 @@ import descriptions from '@/domain/constant/dataSources/dataSourceDescriptions';
...
-            y: formatToMillion(rcsiChartData.rcsi),
-            yRangeMin: formatToMillion(rcsiChartData.rcsiLow),
-            yRangeMax: formatToMillion(rcsiChartData.rcsiHigh),
-          })),
+          dataPoints: countryData.rcsiGraph.map(this.rcsiChartMapper),
         },
       ],
     };
+
+    if (rcsiForcastData) {
```
marinovl7, line 74: check if `process.env.NEXT_PUBLIC_FORECASTS_ENABLED === 'true'`

marinovl7 (approved):
Just one small check to be added, otherwise great!

src/operations/map/FcsAccordionOperations.tsx
marinovl7: Nevermind, I commited the change, great job of yours

PR #184, opened 2025-01-15
feat: change color of the button to check accessibility

ArmanpreetGhotra:
want to test accessibility, that's why need to open draft PR

### DIFF
--- a/src/components/Analytics/CookieConsentPopup.tsx
+++ b/src/components/Analytics/CookieConsentPopup.tsx
@@ -53,7 +53,7 @@ export function CookieConsentPopup() {
         <CustomButton
           color="secondary"
           size="sm"
-          className="-order-1 md:order-2 lg:-order-1 xl:order-2"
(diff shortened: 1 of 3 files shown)


Esoteriker:
Thanks to Arman for the changes! I tested it using Microsoft Edge, and the deployed version achieves a full a11y score of 100. However, Google Chrome still reports a color contrast error. I’m not sure why that’s happening, but the updates Arman made to address the color contrast issue on the analytics button have definitely been effective. @marinovl7 @Tschonti

Tschonti:
I think the 'Allow analytics' button should have a more vibrant color, since we would prefer if the users clicked on that :D

btw I got the 100 score on a11y in chrome (on the deployed version)

ArmanpreetGhotra:
@Tschonti yes we know (we need to change the color), it's just draft pr, so that you and lachezar can check whether this was the problem

Tschonti (approved):
not the biggest fan of this blue as it's not used anywhere else in the app, but it solves the problem, so fine by me

PR #186, opened 2025-01-17
feat: added right margin to theme switch

Lukas0912:
Please check if the space between chatbot button and theme switch looks good to you


marinovl7 (changes requested):
The breakpoint for transforming the navbar into mobile one should be changed. See 1030px overflow issue

Lukas0912:
Alright, I put xl now as a break point. This means it already switches to mobile around 200px earlier than it would have to (considering the issue) but it's the leanest solution. I can also still do something like a custom break at around 1050px but this also means relatively more code compared to the actual change. Have a look and just tell me what you prefer :)

PR #189, opened 2025-01-17
feat: show regions even if there is no region label data

Lukas0912:
Show regions even if there is no label data present


marinovl7:
Conflicts

PR #193, opened 2025-01-18
Feature/f 230 Comparison Portal | Compare Country Regions

jschoedl:
* ~There are some uncaught errors if data is not defined, e.g. for Cambodia (currently fixing it)~
* For now, there are only absolute numbers displayed. The relative numbers are implemented and tested (see `CountryComparisonAccordion`, line 24), but it might be of use to add a switch in the charts toolbar instead of an additional chart.
* The historic line chart does not really make sense if there are very many regions. We might discuss this later. Options to deal with this could be:
  * disabling it with e.g. >15 countries
  * showing aggregate numbers like min, max, mean, median (maybe together with greyish lines of the actual data in the "background")
  * show a warning, but let the user proceed to render the chart if requested

demo:
* Uganda, the country with the highest number of regions (112): https://deploy-preview-193--wfp-hungermap.netlify.app/comparison-portal?tab=region&regionComparisonCountry=253&regions=all
* Comparison of two regions: https://deploy-preview-193--wfp-hungermap.netlify.app/comparison-portal?tab=region&regionComparisonCountry=29&regions=615%2C620
* Comparison of 12 regions: https://deploy-preview-193--wfp-hungermap.netlify.app/comparison-portal?tab=region&regionComparisonCountry=29&regions=all


bohdangarchu:
To-Do: no data available should be shown here. Need to check whether its an issue with our data processing or Categorical Chart
image

bohdangarchu:
Also the selection component started triggering jumps to the top of the page again... This is a known nextUI issue and its a bit annoying 
https://github.com/heroui-inc/heroui/issues/3571

jschoedl:
> To-Do: no data available should be shown here. Need to check whether its an issue with our data processing or Categorical Chart
>image

Fixed that! The data was 0 for some regions which was previously unexpected.

src/components/ComparisonPortal/RegionComparisonAccordion.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * @param {string[] | 'all'} selectedRegions
+ * @param {string | undefined} selectedRegionComparisonCountry
+ */
+export default function RegionComparisonAccordion({
+  selectedRegions,
+  selectedRegionComparisonCountry,
+}: RegionComparisonAccordionProps) {
+  const { data: regionData, isLoading } = useRegionDataQuery(Number(selectedRegionComparisonCountry ?? undefined));
+
+  // TODO (F-254): Toggle this within the chart options. If the pie chart is selected, switch to false and hide the toggle button.
```
marinovl7, line 25: remove such comments if not needed anymore

marinovl7 (approved):

PR #194, opened 2025-01-19
feat: add scroll behavior without using scrollbar

ArmanpreetGhotra:
Now user can scroll the accordions without going to the scrollbar


ArmanpreetGhotra:
@Tschonti thank you for the feedback. i like the way the accordions are looking right now. if i  understood you correctly, above the alerts should accordions appear right?

marinovl7 (approved):

marinovl7:
Maybe this can be fixed in another ticket @ArmanpreetGhotra? Would you like to tackle this as well?

ArmanpreetGhotra:
> Maybe this can be fixed in another ticket @ArmanpreetGhotra? Would you like to tackle this as well?

yes i'll took care of that.

PR #195, opened 2025-01-19
Feature/f 186 multiple tooltips are being shown for countries

Lukas0912:
The issue was that when dragging the map, the cursor could still hover over other countries and trigger the tooltips.
I now disable all other tooltips than the one of the country that the mouse is currently hovering over and enable them again once the dragging is finished. This should solve the other issue.


Tschonti (approved):
niceeee

PR #211, opened 2025-01-24
fix: fix positions of accordions and alerts

ArmanpreetGhotra:


src/components/Accordions/AccordionBoxItems.tsx
```diff
@@ -59,7 +59,7 @@ export default function AccordionBoxItems({
         key={uuid()}
         variant="splitted"
         selectionMode={selectionMode}
-        className="p-0 mb-4"
```
Tschonti, line 62: because of this, the legend accordion (and possibly others) now doesn't have a bottom margin either, which looks weird. maybe this setting could be set via a prop?

Tschonti (changes requested):
The FCS accordion is great, unfortunately the other accordions have also lost their margin

Tschonti (approved):

PR #213, opened 2025-01-27
Feature/f 254 comparison portal of fcs

plaume8:
closes: https://linear.app/world-food-programme-wfp/issue/F-254/comparison-portal-or-percent-of-fcs

@jschoedl @marinovl7 @bohdangarchu 
Please have a look at the comparison portal. Is this what you had in mind?

---
#

image

---
#

image


jschoedl:
Could you maybe disable the pie chart when relative numbers are shown? Usually a pie chart does not make sense then. The rest looks good to me at first glance, thanks!

plaume8:
> Could you maybe disable the pie chart when relative numbers are shown? Usually a pie chart does not make sense then. The rest looks good to me at first glance, thanks!

makes sense!  I ve added it

src/components/Charts/helpers/buttons/ChartRelativeNumbersSwitchButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * This component is tied to the `ChartContainer` and `ChartModal` component and should not be used independently.
+ */
+export default function ChartRelativeNumbersSwitchButton({
+  showRelativeNumbers,
+  setShowRelativeNumbers,
+  size = 4,
+}: ChartRelativeNumbersSwitchButtonProps) {
+  const switchToText = showRelativeNumbers ? 'absolute' : 'relative';
+  const icon = showRelativeNumbers ? (
+    <Scroll className={`h-${size} w-${size}`} />
```
jschoedl, line 18: This won't work for all sizes, since tailwind only loads class names that appear as strings within the files. (Internally it is only using regular expressions to scan the code for class names.) That means that this luckily works if the class name appears somewhere else, otherwise it doesn't.
https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names

(Same thing for PercentageSquare and even ChartSliderButton, which hasn't been changed)

jschoedl: To solve this, you could replace ICON_BUTTON_SIZE in `ChartContainer.tsx` with sth like `ICON_BUTTON_CLASS_NAMES='h-4 w-4'`.

src/components/ComparisonPortal/RegionComparisonAccordion.tsx
jschoedl, line 1: `showRelativeNumbers` should be removed from this file (lines 24-26 and within the `useMemo` dependency array)

jschoedl (changes requested):
The rest LGTM!

src/components/Charts/helpers/buttons/ChartRelativeNumbersSwitchButton.tsx
plaume8: thanks for the catch -> I made it even simpler and replaced the "size: number" attribute of the buttons with a "small: boolean" attribute and let the buttons handle the size selection. I choosed this approach cause the small button components should stay as simple as possible and they are exclusively used by the chart components.

plaume8:
I also added the final icons ("percentage and sigma")

marinovl7 (changes requested):
Merge Conflicts

src/components/Charts/helpers/ChartContainer.tsx
```diff
@@ -104,35 +103,31 @@ export function ChartContainer({
...
-              )
+              !disableDownload && <ChartDownloadButton chartRef={chartRef} chartData={chartData} small={small} />
             }
             {
               // button to trigger the full screen modal; rendered if `disableExpandable` is not selected
               !disableExpandable && (
                 <Tooltip text="Enlarge Chart">
                   <Button isIconOnly variant="light" size="sm" onClick={onOpen}>
-                    <Maximize4 className={`h-${ICON_BUTTON_SIZE} w-${ICON_BUTTON_SIZE}`} />
+                    <Maximize4 className={`h-${small ? 3 : 4} w-${small ? 3 : 4}`} />
```
jschoedl, line 130: This won't make any difference, but technically, `small ? 'h-3 w-3' : 'h-4 w-4'` would be better here as it includes the actual classname again. :)

src/components/Charts/helpers/buttons/ChartDownloadButton.tsx
```diff
@@ -10,13 +10,15 @@ import ChartDownloadButtonOperations from '@/operations/charts/ChartDownloadButt
...
+export default function ChartDownloadButton({ chartRef, chartData, small = false }: ChartDownloadButtonProps) {
+  const SIZE = small ? 3 : 4;
+
   return (
     <Popover placement="bottom" offset={10} backdrop="opaque">
       <PopoverTrigger>
         <Button isIconOnly variant="light" size="sm">
           <Tooltip text="Export Chart / Data" offset={20}>
-            <DocumentDownload className={`h-${size} w-${size}`} />
+            <DocumentDownload className={`h-${SIZE} w-${SIZE}`} />
```
jschoedl, line 21: Same here.

src/components/Charts/helpers/buttons/ChartRelativeNumbersSwitchButton.tsx
```diff
@@ -11,13 +11,14 @@ import { ChartRelativeNumbersSwitchButtonProps } from '@/domain/props/ChartConta
...
+  small = false,
 }: ChartRelativeNumbersSwitchButtonProps) {
+  const SIZE = small ? 3 : 4;
   const switchToText = showRelativeNumbers ? 'absolute' : 'relative';
   const icon = showRelativeNumbers ? (
-    <Scroll className={`h-${size} w-${size}`} />
+    <Sigma className={`h-${SIZE} w-${SIZE}`} />
   ) : (
-    <PercentageSquare className={`h-${size} w-${size}`} />
+    <Percent className={`h-${SIZE} w-${SIZE}`} />
```
jschoedl, line 21: And here, you could also use `CLASS_NAMES = ...` to avoid repeating sth

src/components/Charts/helpers/buttons/ChartSliderButton.tsx
```diff
@@ -18,7 +20,7 @@ export default function ChartSliderButton({ showSlider, setShowSlider, size = 4
           setShowSlider(!showSlider);
         }}
       >
-        <Settings className={`h-${size} w-${size}`} />
+        <Settings className={`h-${SIZE} w-${SIZE}`} />
```
jschoedl, line 23: ... and here.

jschoedl (changes requested):
Doing it with the `small` prop seems good, but same thing again: you still did not include the full classnames in your code. Probably `w-3 w-4 h-3 h-4` all appear somewhere else in the code, which is why they are known to Tailwind and work. But in case there is a change to the repo and there would no longer be any file including e.g. `h-3` as single string, Tailwind would think that this class is never use and would exclude it from the generated CSS files -> the class would no longer have an effect. That's why it's the best practice to always include Tailwind classnames as they are, not as template strings.

plaume8:
> Doing it with the `small` prop seems good, but same thing again: you still did not include the full classnames in your code. Probably `w-3 w-4 h-3 h-4` all appear somewhere else in the code, which is why they are known to Tailwind and work. But in case there is a change to the repo and there would no longer be any file including e.g. `h-3` as single string, Tailwind would think that this class is never use and would exclude it from the generated CSS files -> the class would no longer have an effect. That's why it's the best practice to always include Tailwind classnames as they are, not as template strings.

@jschoedl thanks for explaining it once again -> now I finally understand it! -> i changed all relative tailwind sizings to `small ? 'h-3 w-3' : 'h-4 w-4'`

plaume8:
@marinovl7 merge conflicts are resolved

PR #216, opened 2025-01-27
Feature/f 250 query parameter copy link

ArmanpreetGhotra:
features:
* Introduced a `QueryParamsContext`
  * has one single state for all query params
  * provides setters and getters for the other map contexts below (mapType, alert, countryId)
  * in theory it would be useful to include the comparison and wiki params as well -> maybe we can note that somewhere?
* added **query params for the map** 
* added a **share button** to copy /share the current URL from the map
* if Vegetation / Rainfall is selected, zoom out


jschoedl:
Hi, if anyone has time, please check/test this. Arman is at work today, so I took over the rest. I know that these changes are pretty complex for the last day of coding, sorry for that.

I'll write a short documentation on how to add query params after these changes Afterwards I would be available from ca. 21:30 on if something still needs to be adapted.

bohdangarchu (changes requested):
Wow, great job! Two things
1. Query param update is a bit slow, not sure if we can do something about this tho
2. I noticed that reloading or opening a link with query params can lead to a really buggy page (see video). This is very weird and also not sure if this can be fixed so short term
Here is the video (its too big for github)
https://we.tl/t-uHXKuztlad

jschoedl:
1. Yeah, query params are always slow for some reason, but the state values itself (i.e. the UI updates) should change immediately.
2. Hm, I haven't ever seen this actually. Not sure why it happens

jschoedl:
Ah, if we leave it out, the (already merged) PR to the docs does not make sense anymore.

PR #221, opened 2025-01-28
Fix multiple tooltips bugs after dragging map

Lukas0912:
I found a imo better solution for the bug with tooltips that are not removed after dragging. I implemented it for all modes in global and country view. I will update the docs of the MapOperations file as well again


marinovl7 (changes requested):
Great, thanks for solving this, can you also please resolve the merge conflicts and merge the PR

PR #222, opened 2025-01-28
use new api endpoints

Tschonti:
Should not be merged until the new version of the backend API is deployed. 

changes:
- the countryData method now uses the optimized v3 endpoint 
- the workaround is removed from the alertsAndCentroids method as the endpoint was fixed on the backend
- the adm0data is fetched in two chunks
- the forecasted data is fetched from the backend now. A new envvar, `NEXT_PUBLIC_FORECAST_API_URL` has been introduced for its URL as it might be different to the previous apis.
- small refactor in the CSP middleware because there are a lot of URLs
- the name of the country is now retrieved from the new FCS endpoint in the HungerAlert


src/infrastructure/repositories/GlobalDataRepositoryImpl.ts
```diff
@@ -43,11 +43,14 @@ export default class GlobalDataRepositoryImpl implements GlobalDataRepository {
...
+    const response1 = await fetch(`${process.env.NEXT_PUBLIC_V3_API_URL}/adm0data1.json`, {
+      next: { revalidate: 3600 * 12 },
     });
-    const data: ResponseWrapper<CountryMapDataWrapper> = await response.json();
-    return data.body;
+    const response2 = await fetch(`${process.env.NEXT_PUBLIC_V3_API_URL}/adm0data2.json`, {
+      next: { revalidate: 3600 * 12 },
+    });
+    const [data1, data2]: CountryMapDataWrapper[] = await Promise.all([response1.json(), response2.json()]);
+    return { type: data1.type, features: [...data1.features, ...data2.features] };
```
marinovl7, line 53: This is wrong, you are awaiting the responses sequentially, but exectuing promise.all on response.json(). 

**Suggestion**

```typescript
async getMapDataForCountries(): Promise<CountryMapDataWrapper> {
  const [response1, response2] = await Promise.all([
    fetch(`${process.env.NEXT_PUBLIC_V3_API_URL}/adm0data1.json`, {
      next: { revalidate: 3600 * 12 },
    }),
    fetch(`${process.env.NEXT_PUBLIC_V3_API_URL}/adm0data2.json`, {
      next: { revalidate: 3600 * 12 },
    })
  ]);

  const [data1, data2]: CountryMapDataWrapper[] = await Promise.all([
    response1.json(),
    response2.json()
  ]);

  return {
    type: data1.type,
    features: [...data1.features, ...data2.features]
  };
}
```

marinovl7 (changes requested):
What about the modified fcs endpoint? Regarding country names which hasn't been here in the previous version

marinovl7 (approved):
Well done!

PR #224, opened 2025-01-29
feat: compare button

bohdangarchu:


marinovl7 (changes requested):
Please add tooltips to the buttons if they are only icons. The same should be done for expand sidebar button, country insights button, chatbot button text should change. I know they are not part of this PR, but can you please add them when adding the others?

PR #226, opened 2025-01-29
feat: add absolute positioned popover outside of the accordion

georgi4444:
The only solution I can think of.

Works but it doesnt look good, there is some weird flashing


georgi4444:
Actually I think I found another way

georgi4444:
Better solution found: https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/228

PR #238, opened 2025-02-01
feat: disable top-level map tooltips for ios devices (IOS Tooltips Option 1)

georgi4444:
Disable hovering on ios on the top level layers


georgi4444:
Another option to solve the issue: https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/239

PR #239, opened 2025-02-01
feat: Add small addition info on iOS devices "tap again for details" (iOS Tooltips Option 2)

georgi4444:
Add small addition info on iOS devices "tap again for details"
IMG_9076
IMG_9077


georgi4444:
Another option to solve the issue: https://github.com/jst-seminar-rostlab-tum/wfp-hunger-map/pull/238

PR #241, opened 2025-02-06
fix: missing country name

bohdangarchu:


src/components/ComparisonPortal/CountryComparisonAccordion.tsx
```diff
@@ -59,6 +59,10 @@ export default function CountryComparisonAccordion({
     [countryDataQuery, countryIso3DataQuery]
   );
 
+  useEffect(() => {
```
plaume8, line 62: Looks good to me and works fine on my machine after several tests ... I would just add a small comment explaining why it is necessary.

